PR #63, opened 2013-11-14
Improving multi distro support.

ppouliot:
This improves deployment on multiple distributions.  Slave deployment on ubuntu systems using newer bits is now supported.


jhoblitt:
:-1: for this PR in its present form.  The history needs to be squashed and the merge commits removed.  The
dependency being added on  `openstack-hyper-v/windows_common` is for a module not published on the forge -- which would make this module uninstallable via PMT.  I object to the inclusion of the `jenkins-slave.exe` binary, even though it is small.  I am concerned about the way `jenkins.model.JenkinsLocationConfiguration.xml` is being managed since `replace` is not `false` and the template can not be overridden.  It also goes against the pattern this module has of managing global configuration via the cli jar.  Debian/RHEL init scripts are already present.

jhoblitt:
@ppouliot I'm going to close this PR due to the issues I noted in my previous comment.  I would be happy to work with you in the future on adding windows support.

PR #138, opened 2014-05-29
add ability to disable a plugin

### LINKED ISSUE #136: support for disabling plugins
I'd love to see support for disabling Jenkins plugins. The core distribution ships with some plugins that aren't necessarily used in all deployments (i.e. CVS). This can be handled by writing an empty "{plugin-name}.jpi.disabled" file to the plugins directory.

jlambert121:
Resolves #136


jhoblitt:
This functionality was introduced by #209 . This PR can be closed.

PR #164, opened 2014-08-15
Ensure plugin download exec is idempotent

davidsnyder:
Since this is a blind exec, it's re-downloading the plugins every puppet run, which then triggers the jenkins service notify.  This means that jenkins is needlessly restarted every time puppet agent runs, killing any currently running jobs.  This skips the exec if the file already exists (only for .hpi)


jhoblitt:
I see that the cli has the ability to do plugin management, probably using the plugin manager exposed through the http interface (haven't checked the source).  Does it not do something that's being done by `puppet::plugin`?

jhoblitt:
@davidsnyder We do appreciate your efforts! This issue should be resolved by #336, so I'm going to close out this pR.

PR #225, opened 2014-12-10
Update jenkins-slave.RedHat

telegrapher:
Add small space to the final command line.
Without it, it doesn't work in CentOS 5.


jhoblitt:
It looks like the PR branch has had a number of unrelated changes merged into it well beyond the original intent of this PR and the the history is extremely dirty. :-1:

jhoblitt:
I suspect that #314 resolves the original intent of this PR so I'm going to close it.

Issue #226, opened 2014-12-16
Support for setting UUID of user during credentials creation


madAndroid:
When I create a job using `jenkins::job`, with a specified UUID in the template, I'd like to be able to set that UUID during credential creation.

ltutar:
For my own understanding.
Do you mean the 'id' line as shown below? I can set this value through Jenkins GUI but not through puppet? and breaks my automated jenkins provisioning.

    <java.util.concurrent.CopyOnWriteArrayList>
        <com.cloudbees.plugins.credentials.impl.UsernamePasswordCredentialsImpl>
          <scope>GLOBAL</scope>
          <id>dbd9a6c8-c7c5-428b-946e-9c3e2333ef58</id>
          <description>vagrant user for git</description>
          <username>vagrant</username>
          <password>w5oFlgURdduq2RlLJQBJUTmOEeJvFdC4geAHiW3K0Hk=</password>
        </com.cloudbees.plugins.credentials.impl.UsernamePasswordCredentialsImpl>
      </java.util.concurrent.CopyOnWriteArrayList>

madAndroid:
@ltutar - yes indeed  ... it's used in numerous places, for example:
http://docs.openstack.org/infra/jenkins-job-builder/scm.html#scm.git refers to it as 'credentials-id', which allows you to set it for use with JJB

I also use it in the job-dsl when specifying credentials to use when needed

PR #227, opened 2014-12-16
Add parameter to set user uuid in jenkins::credentials define

madAndroid:
This addresses #226 - It's largely influenced by the Opscode jenkins module, which has this functionality - https://github.com/opscode-cookbooks/jenkins/blob/master/libraries/credentials.rb

### DIFF
--- a/files/puppet_helper.groovy
+++ b/files/puppet_helper.groovy
@@ -162,7 +162,7 @@ class Actions {
   /////////////////////////
   // create credentials
   /////////////////////////
-  void create_or_update_credentials(String username, String password, String description="", String private_key="") {
+  void create_or_update_credentials(String username, String password, String id="", String description="", String private_key="") {
     def global_domain = Domain.global()
     def credentials_store =
       Jenkins.instance.getExtensionList(
@@ -170,10 +170,13 @@ class Actions {
       )[0].getStore()
 
     def credentials
+    if (id == "") {
+      id = null
+    }
     if (private_key == "" ) {
       credentials = new UsernamePasswordCredentialsImpl(
         CredentialsScope.GLOBAL,
-        null,
+        id,
         description,
         username,
         password
@@ -187,7 +190,7 @@ class Actions {
       }
       credentials = new BasicSSHUserPrivateKey(
         CredentialsScope.GLOBAL,
-        null,
+        id,
         username,
         key_source,
         password,
--- a/README.md
+++ b/README.md
@@ -301,6 +301,24 @@ private key:
     }
 ```
 
+*_Setting a UUID:_*
+
+You can also specify a UUID to use with the credentials, which will be used to
+identify the credentials from within the job config. This is necessary when setting
+credentials for use with the [git plugin](http://docs.openstack.org/infra/jenkins-job-builder/scm.html#scm.git), for example.
+
+You can either manually generate a UUID from a site like https://www.uuidgenerator.net,
+or use the UUID from an existing user, which is accessible within the URL of the
+Jenkins console when managing an existing user's credentials.
+
+```puppet
+    jenkins::credentials { 'deploy-user':
+      password            => '',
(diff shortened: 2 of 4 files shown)


madAndroid:
@rtyler - Any progress on testing this? :) Please let me know if more tests are required

xiii:
@rtyler @madAndroid that would be very useful! hope it gets merged soon!

jhoblitt:
The last CI run failed for this PR and it needs to be rebased on the current master due to changes introduced by #285.

madAndroid:
Cool, no worries - I'll rebase and sort out any conflicts

madAndroid:
Rebased, and tests fixed :) ( https://travis-ci.org/jenkinsci/puppet-jenkins/builds/63873665 )

files/puppet_helper.groovy
```diff
@@ -170,10 +170,13 @@ class Actions {
       )[0].getStore()
 
     def credentials
+    if (id == "") {
+      id = null
+    }
```
jhoblitt: From looking at the existing code, I thought that the cli argument to method call conversion, `actions."$action"(*args[1..-1])`, must be requiring string params to be defaulted to an empty string.  However, I just tested it and there doesn't appear to be an issue with defaulting a string param to `null`.  

The `id` param could be defaulted to `null` and then this check wouldn't be needed.

jhoblitt: Actually, that might not make any sense.  I'm not sure how you would pass null from the cli.

madAndroid: > I'm not sure how you would pass null from the cli.

That's pretty much the issue I came up against, which is why I ended up taking this route.

jhoblitt:
:+1: This looks good to me!

CI didn't run; travis has been acting strange the last few days.  I've had to force push some of my own branches in order to get it trigger.

madAndroid:
Odd, I did see the run on travis after I pushed the latest changes up .. results are here, AFAICT ->  https://travis-ci.org/jenkinsci/puppet-jenkins/builds/63873665

madAndroid:
An example of the use case where you may want to set the UUID during credential creation: [JJB's SCM plugin](http://docs.openstack.org/infra/jenkins-job-builder/scm.html#scm.git) allows you to set `credential-id` field, which allows you to have a predefined UUID for easy rebuild of envs

madAndroid:
Is there anything more I can do to have this request merged? ( @jhoblitt / @rtyler )

jhoblitt:
This looks pretty good to me.  Mentioning the new param in the README might not hurt. (eg., https://github.com/jenkinsci/puppet-jenkins#credentials)

One thing of note is that this does change the order of params to `create_or_update_credential` -- I'm not sure if that's ever been considered a public API, @rtyler ?

madAndroid:
@rtyler - anything more I need to do to have this merged? I've added tests, would you like me to expand upon those?

madAndroid:
Hi, is there anything further than can be done to have this merged? Do you need me to justify the reason for the addition, and perhaps explain and describe our use case? 

@jhoblitt - in response to your query: the ordering of args in `create_or_update_credential` is taken directly from the credentials-plugin source: https://github.com/jenkinsci/credentials-plugin/blob/master/src/main/java/com/cloudbees/plugins/credentials/impl/UsernamePasswordCredentialsImpl.java#L56-L64
ie:
```
    /**
     * Constructor.
     *
     * @param scope       the credentials scope
     * @param id          the ID or {@code null} to generate a new one.
     * @param description the description.
     * @param username    the username.
     * @param password    the password.
     */
```
I've rebased against the latest HEAD of master, and fixed tests that changed as a result of another merge.

rtyler:
@madAndroid nope, thanks for the bump and the good contribution :)

madAndroid:
Awesome, thanks! :-)

PR #228, opened 2014-12-17
Few fixes to work with Centos 7

filipgorny:
I have struggled with errors cause by those lines so I tought I would share as a pull request.
First one is about java when java executable is not in the path.
Second one is return code 4 when deleting non existing file with rm -Rf

### DIFF
--- a/manifests/cli.pp
+++ b/manifests/cli.pp
@@ -29,7 +29,7 @@ class jenkins::cli {
   $port = jenkins_port()
 
   # The jenkins cli command with required parameter(s)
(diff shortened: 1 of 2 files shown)


jhoblitt:
I am not seeing an issue with the java path on centos 7.0 and I've been testing fairly heavily over the last month.

jhoblitt:
@filipgorny The exec resources are no long present in `jenkins::plugin`.  I've been unable to reproduce java path issues on el7 so I'm going to close out this PR.  If you are still able to reproduce this, please let us know.

PR #231, opened 2015-01-09
Add ldap and roles

abayer:
So this is very very preliminary and annoyingly chock full of my IDE reindenting the shit out of things, but...

* Rework puppet_helper.groovy to use groovy.util.CliBuilder, so that we don't have to worry about things like order of arguments and the like. Not huuuuuge yet, but I think it's valuable, so that we could do null values for arguments etc.
* As a result, change all the methods to take a Map as their only argument - this is so that the magic method dispatch stuff doesn't need to know the names of the arguments for each method. That map is generated from CliBuilder's results (with some twiddling to always have unspecified arguments show up as an empty string, rather than null, just to make things safer), and then the values there are accessed by "args.whatever_argument_flag_name".
** The above two may be excessive, I admit. But future proofing, yo. Future proofing.
* In security.pp, add ldapconfig and permissions variables, intended to be specified in hiera. This stuff is the gnarly part - to get hashes in particular from Puppet to something puppet_helper.groovy can know about is...yeah. So we serialize these down to /tmp/something.json via templates, and then pass the path to those files as arguments to puppet_helper. puppet_helper then knows to slurp them in and use 'em. Woo.
* And then we take the new possible values for security_model (global_matrix) and user_realm (ldap) - if they're selected, we instantiate/populate the objects using the data we just deserialized. Tada!

So yeah, no idea if it actually *works* yet. But I wanted to throw this up to make sure I wasn't too batshit. As I write this, I'm thinking I probably am in some parts (most notably the CliBuilder and args as Maps bits - I'm not 100% sure we actually *need* either of those things as this currently stands, but they both come in handy for further extensions in the future). So tell me what's too crazy and I will clean it out.

### DIFF
--- a/files/puppet_helper.groovy
+++ b/files/puppet_helper.groovy
@@ -34,6 +34,7 @@ import hudson.security.FullControlOnceLoggedInAuthorizationStrategy
 import hudson.security.GlobalMatrixAuthorizationStrategy
 import hudson.security.HudsonPrivateSecurityRealm
 import hudson.security.LDAPSecurityRealm
+import hudson.security.Permission
 import hudson.util.Secret;
 import jenkins.model.*
 import jenkins.security.ApiTokenProperty;
@@ -98,7 +99,10 @@ class Actions {
 
     cli._(longOpt:'user_realm', args:1, "Used in set_security")
 
-    cli._(longOpt:'ldap_config_file', args:1, "JSON file containing LDAP configuration")
+    cli._(longOpt:'ldap_config_file', args:1, "JSON file containing LDAP configuration. Required for ldap user_realm.")
+
+    cli._(longOpt:'matrix_auth_config_file', args:1, "JSON file containing array of permission strings. Required for global_matrix security model.")
+
     def opts = cli.parse(cliArgs)
 
     def optMap = [:]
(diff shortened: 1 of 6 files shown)


adilmehmoodbutt:
How can i  migrate Gitlab accounts from standard accounts to ldap users along with work history.

rtyler:
This is something I'd love to get @mwedgwood-rmn's eyes on if he has time for it too

mwedgwood-rmn:
I definitely like the changes. I'm completely unsurprised that someone has vastly improved upon what I wrote :)

Looks like it needs a rebase (at least), but otherwise :+1: and thanks @abayer

rtyler:
@abayer gimme a rebase d00d. Also, maybe lie to me and tell me you've tested it thoroughly first :)

eddyhub:
I am new to jenkins but with "java -jar /usr/share/jenkins/jenkins-cli.jar -s http://127.0.0.1:8080 groovysh" I get a shell which tells me, that I can use groovy 1.8.9 (terminal output: "Groovy Shell (1.8.9, JVM: 1.7.0_75)"). In http://docs.groovy-lang.org/docs/groovy-1.8.9/html/api/groovy/util/package-summary.html I can't find a class named CliBuilder, so how is it possible that this script works.

tedivm:
@abayer, could you rebase this so people can give it a good look?

jhoblitt:
@abayer Could you split the conversion to using the Cli builder and the new features into separate PRs/commits?  There's a lot going on in this branch and there's no test coverage of the helper script.

jhoblitt:
@abayer This PR needs a rebase and ideally to be split up into multiple PRs.  Are you still interested in getting this merged?

boydj:
Any chance of this getting rebased and merged? I'd do it but I'm not entirely sure of how to rebase somebody else's PR.
@abayer?

jhoblitt:
@boydj What I typically do is pull the branch the PR is based on, do the rebase, push, open a new PR with "closes #XXX" in the PR commit message.

jhoblitt:
@boydj Note the request to split this into two PRs.  https://github.com/jenkinsci/puppet-jenkins/pull/231#issuecomment-119001400 In addition, it would be less intrusive to get this working with the new `jenkins_security_realm` and `jenkins_authorization_strategy` types.

PR #233, opened 2015-01-10
fixes timeouts on restart

endyman:
Puppet modules fails while modifying users, credentials or security when the server is not yet up. This PR implements timeouts.


endyman:
travis issues are caused by bundler not by this PR

rtyler:
I think increasing the timeouts is okay, but there's a lot of repetition of hard-coded timeouts and tries here, can you refactor those values into `$::jenkins::params`? That might make this easier for people to override in the future too

endyman:
done!

PR #234, opened 2015-01-12
added user_hash 

WianVos:
i've taken the liberty of adding a user_hash parameter to the init.pp so it has the same capabilities as plugins and jobs. 

I've also added rspec tests and updated the inline docs .. 

plz merge


jhoblitt:
Just FYI - I have a working type/provider for jenkins user management but it needs some minor refactoring and a history cleanup before its ready for merge.

ricoli:
I was about to add the user_hash too, just noticed this pull request. This is much needed, and I'll likely add the same for credentials.

.gitignore
```diff
@@ -11,3 +11,4 @@ README.pdf
 spec/fixtures
 Gemfile.lock
 .vagrant*
+.idea*
```
ricoli: not related

ricoli:
Actually this has already been done and merged in https://github.com/jenkinsci/puppet-jenkins/pull/297

PR #235, opened 2015-01-14
Make creation of user optional

bodepd:
Currently, the creation of jenkins user
and group is protected by if !defined. This
can be extremely annoying in cases where code
is trying to install a competing jenkins user/group.

In my specific case, I'm blocked by this code
leading to duplicate jenkins users (and I need
my definition to be created first). In general,
I thought that these kinds of issue were resolvable
via enusre that classes were including in a certain
order, but I cannot seem to resove it in my specific use
case which involves:
* nested class includes
* serialized user resources from hiera
* create_resources

### DIFF
--- a/manifests/plugin.pp
+++ b/manifests/plugin.pp
@@ -7,7 +7,7 @@
 #   create this content from a template or any other mean.
 #
 # update_url = undef
-#   
+#
 define jenkins::plugin(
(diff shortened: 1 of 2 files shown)


bodepd:
These failures do not seem to have anything to do with my patch.

PR #244, opened 2015-01-31
Add instructions for acceptance tests

petems:
Lets make @rtyler happy! :smile:


rtyler:
This doesn't look like beer...

PR #246, opened 2015-02-08
add firewall module as a dependency

cristi1979:
also fix versions: x.y is not a valid version, x.y.z. is (according to gepetto)

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -19,9 +19,10 @@
     }
    ],
   "dependencies": [
-    { "name": "puppetlabs/stdlib", "version_requirement": ">= 2.0.0 < 5.0" },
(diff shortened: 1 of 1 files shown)


rtyler:
@matthewbarr if I remember correctly we explicitly decided against this dependency correct?

jchristi:
Please don't add this. Some organizations like mine have their own custom firewall module (for better or for worse).  Adding this dependency would result in a namespace conflict.

rtyler:
@christ66 I'm going to close this, I'm in agreement with @jchristi and our previous decision regarding keeping the puppetlabs-firewall dependency optional.

christ66:
@rtyler I think you mean @cristi1979 :smile:

matthewbarr:
Yep, explicitly not added :)

PR #255, opened 2015-03-18
Set ubuntu default directory to /etc/default

iszak:
Currently the path for Jenkins slave config on Ubuntu is ```/etc/sysconfig``` which is the default which doesn't exist in the base installation, to work around this I created a file resource which the class depended on;
```
file { '/etc/sysconfig':
  ensure => 'directory'
}

class { '::jenkins::slave':
  require => File['/etc/sysconfig']
}
```
It would be nice if it used the same as the Debian installation path.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -15,7 +15,7 @@ class jenkins::params {
   $cli_try_sleep         = 10
 
   case $::osfamily {
-    'Debian': {
+    'Debian', 'Ubuntu': {
       $libdir = '/usr/share/jenkins'
     }
     default: {
--- a/manifests/repo.pp
+++ b/manifests/repo.pp
@@ -20,7 +20,7 @@ class jenkins::repo {
           Anchor['jenkins::repo::end']
       }
 
-      'Debian': {
+      'Debian', 'Ubuntu': {
         class { 'jenkins::repo::debian': }
         Anchor['jenkins::repo::begin'] ->
           Class['jenkins::repo::debian'] ->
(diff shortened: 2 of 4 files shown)


manifests/params.pp
```diff
@@ -15,7 +15,7 @@
   $cli_try_sleep         = 10
 
   case $::osfamily {
-    'Debian': {
+    'Debian', 'Ubuntu': {
```
rtyler: I don't believe this is necesasary, on my Ubuntu 12.04 LTS the `osfamily` fact reports "Debian", not "Ubuntu"

rtyler:
I'm not sure this change is necessary

```
tyler@artichoke:~$ facter osfamily
Debian
tyler@artichoke:~$ cat /etc/issue
Ubuntu 12.04.4 LTS \n \l

tyler@artichoke:~$ 
```

jhoblitt:
Under both ruby facter and cfacter, Ubuntu is part of osfamily Debian.  Eg., https://github.com/puppetlabs/facter/blob/ececbe6b3cfbad162400bbd58f81ba84b0e4bafb/lib/src/facts/resolvers/operating_system_resolver.cc#L271

iszak:
Yes, this certainly seems the case I think I was getting it confused with operatingsystem fact, but in any case the reason I did this change was it wasn't installing the daemon package with puppet 3.7.x with future parser and Ubuntu 12.04, I will revisit this and see if I can narrow it down.

iszak:
I believe the comments are right so I will be closing this, I think it was related to issues with my facts.

PR #257, opened 2015-03-23
Support getting Jenkins .xml job descriptions from Puppet fileserver

gdhbashton:
Jenkins-job-builder is too brittle for me so I have returned to simply storing the full `config.xml` jobs in my Puppet repo.

If I use `jenkins::job_hash` as-is, it would require the XML raw job to be pasted into hiera config which is difficult to read. This PR adds a simple 'sourced' condition to get the module to get the config from a puppet filie server.. e.g. this permits the following cleaner syntax in a hiera conf:

```
jenkins::job_hash:
  genersys-outgoing-promotion-file-writer:
    config: 'itv_genersys_data/jenkins/qa1/genersys-outgoing-promotion-file-writer.xml'
    sourced: 1
  genersys-outgoing-promotion-ftp-poller:
    config: 'itv_genersys_data/jenkins/qa1/genersys-outgoing-promotion-ftp-poller.xml'
    sourced: 1
```

Cheers,
Gavin

### DIFF
--- a/manifests/job.pp
+++ b/manifests/job.pp
@@ -20,6 +20,7 @@ define jenkins::job(
   $config,
   $jobname  = $title,
   $enabled  = 1,
+  $sourced  = 0,
   $ensure   = 'present',
 ){
 
@@ -29,7 +30,11 @@ define jenkins::job(
     }
   } else {
     jenkins::job::present { $title:
-      config  => $config,
+      if ($sourced) {
+        config => file($config),
+      } else {
+        config  => $config,
+      }
       jobname => $jobname,
       enabled => $enabled,
     }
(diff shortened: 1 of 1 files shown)


manifests/job.pp
```diff
@@ -20,6 +20,7 @@
   $config,
   $jobname  = $title,
   $enabled  = 1,
+  $sourced  = 0,
```
jhoblitt: This should be a proper boolean instead of an integer.  However, I believe a `source` and/or `content` parameter in the style of the core `file` resource would be more idiomatic.

rtyler:
The test failures look legitimate here, and I would love to see a test case for the `sourced` branch before pulling this in

gdhbashton:
OK cool - thanks for the feedback. I'll bash the code into a better shape and see if I can scare up some tests.

gdhbashton:
Please see PR #290 since the repo this PR was based on has now been deleted. Cheers, Gavin.

PR #272, opened 2015-05-02
use mainline puppetlabs_spec_helper gem

jhoblitt:
The 'jenkins-infra/puppetlabs_spec_helper' fork is significantly behind the
mainline gem release.


rtyler:
If all the tests are passing, then sure.

PR #284, opened 2015-05-03
remove puppet > 3 < 3.6 from travis matrix

jhoblitt:
To reduce wall clock CI execution time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,11 +10,6 @@ script:
   - "bundle exec rake spec SPEC_OPTS='--format documentation'"
 env:
   - PUPPET_VERSION="~> 2.7.0"
-  - PUPPET_VERSION="~> 3.1.0"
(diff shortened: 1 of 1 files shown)


rtyler:
I don't think that Puppet 2.7 should be removed from the matrix. There's still lots of folks out there who haven't made the jump to 3 and I'd like to make sure we still support them.

jhoblitt:
2.7 is completely excluded from the matrix as it is only compatible with ruby 1.8.7.  As is, its just dead configuration.

rtyler:
Huh, I hadn't realized that. I suppose I'll have to add 1.8.7/2.7 in a later commit to support all my EoL buddies out there :P

jhoblitt:
I'll drop that commit from the PR momentarily.

jhoblitt:
Its gone.

PR #285, opened 2015-05-03
convert raw execs of puppet_helper.groovy to jenkins::cli::exec define

jhoblitt:


rtyler:
Hooray abstractiosn

jhoblitt:
Indeed.  :)

Hopefully, most of these will get replaced with a native type.  I wrote jenkins::cli::exec when I was trying jenkins for the first time (at the end of march) and hadn't yet appreciated the complexities of dealing with the cli jar or some of the obstacles in the jenkins API.  Implementing types and providers turned out to be much more difficult than I originally estimated.  I've essentially finished a jenkins_user resource type but want to do some refactoring before PRing it.

PR #290, opened 2015-05-06
Support getting external .xml job descriptions

gdhbashton:
This is a new PR to replace #257 whose repo has now been deleted and replaced. The new code should be a little more elegant, and now contains rspec tests.

Cheers,
Gavin.

### DIFF
--- a/manifests/job.pp
+++ b/manifests/job.pp
@@ -22,7 +22,7 @@
 #
 define jenkins::job(
   $config,
-  $source = '',
+  $source   = undef,
   $jobname  = $title,
   $enabled  = 1,
   $ensure   = 'present',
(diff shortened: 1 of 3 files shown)


spec/defines/jenkins_job_spec.rb
```diff
@@ -1,5 +1,8 @@
 require 'spec_helper'
 
+Puppet::Util::Log.level = :debug
+Puppet::Util::Log.newdestination(:console)
```
jhoblitt: Why is this needed?

gdhbashton: It's not needed at all. It's debugging code that didn't get removed before I pushed....

manifests/job.pp
```diff
@@ -18,6 +22,7 @@
 #
 define jenkins::job(
   $config,
+  $source = '',
```
jhoblitt: Why not make the default `undef`?  I suggest validating the param with `validate_string()`.

manifests/job.pp
```diff
@@ -28,10 +33,18 @@
...
+        config  => $config,
+        jobname => $jobname,
+        enabled => $enabled,
+      }
+    } else {
+      jenkins::job::present { $title:
+        config  => file($source),
+        jobname => $jobname,
+        enabled => $enabled,
+      }
```
jhoblitt: Instead of duplicating the resource, I would suggest using a selector to set a variable for the `config` param.

PR #292, opened 2015-05-06
Feature/puppet helper util

jhoblitt:


jhoblitt:
This is essentially a no-op but its foundational for planned PRs that I'm attempting to break down into easily digested patch sets.

jhoblitt:
I've amended  and repushed one of the commit messages to note that this is a breaking change to the output of user_info(). There are no in-tree consumers of this method so, AFAIK, it should be safe.

PR #302, opened 2015-05-18
Package name no longer hardcoded

ricoli:
It was hardcoded in at least these two occurrences.


matthewbarr:
LGTM.  Any one else have any concerns?

PR #303, opened 2015-05-18
Puppet helper slaveagentport

rmestrum:
* add get_slaveagent_port()/set_slaveagent_port() methods to puppet_helper
* add slaveagentport param to jenkins class


matthewbarr:
You'd also want to add this to the documentation- at the very least to the init.pp documentation. All parameters need to be declared, and explained at the top of init.p

PR #304, opened 2015-05-18
Add "basic" support for OpenBSD. 

buzzdeee:
Support only "basic", since a few things are untested, and known to not work (yet).

What works and is tested:
- installation via package
- jenkins service management
- plugins_string fact and therefore plugin management

What is definately not working:
- jenkins::cli (jar is not in the PATH), I have to "enhance"
  puppetlabs/java before that, to add a fact pointing to the right
  directory, or add another variable to params.pp ;)

Otherwise untested:
- jobs management
- jenkins slave configuration

Some changes had to be done to the params.pp. Some information that
were hard-coded before, are now OS dependent set there. For all the
other OS, kept as it was there already, or somewhere else hard-coded.

On OpenBSD, and other *BSD, there is no group 'root', therefore, whereever
access permissions for a 'root' group were made, changed it to the gid of '0'

The jenkins system user and group are '_jenkins', therefore added these
as variables. Jenkins users home directory is /var/jenkins, instead of the
otherwise hardcoded /var/lib/jenkins, and further, jenkins is installed
in /var/lib/jenkins/.jenkins.

Because of above facts, most changes to manifests/plugins.pp, because the
plugin_parent_dir assumption, that this is the same as the jenkins users
home directory is not true on OpenBSD. Also, wget is available as package,
but installed in /usr/local/bin. Therefore added this to the path for the command.

Further, made repo management OS dependent, and for OpenBSD, set the value
to false.

Services on OpenBSD get configured via flags to the OpenBSD service provider.
Since I had seen problems with older Puppet versions, manifests/services got
a little bit enhanced, to only make use of that flag, when on OpenBSD.

Last but not least, a bunch of specs added here and there, to reflect OpenBSD.

Tested on OpenBSD amd64 with Puppet 3.7.4, jenkins-1.605 and jenkins-1.596.1, that is all available as packages.

haven't added OpenBSD to metadata.json, due to above mentioned bits that are not (yet ;) supported.

Hope that all that makes sense, and can be merged. Let me know if there is something that needs clarification, or changes or whatnot, and I'll be happy to work on it.

cheers.

### DIFF
--- a/lib/puppet/jenkins.rb
+++ b/lib/puppet/jenkins.rb
@@ -4,7 +4,11 @@ module Puppet
     # @return [String] Full path to the Jenkins user's home directory
     def self.home_dir
       begin
-        return File.expand_path('~jenkins')
+        if Facter.value(:osfamily) == 'OpenBSD'
+          return File.expand_path('~_jenkins')
+        else
+          return File.expand_path('~jenkins')
+        end
       rescue ArgumentError
         # The Jenkins user doesn't exist!
         return nil
@@ -13,7 +17,11 @@ module Puppet
 
     # @return [String] Full path to the Jenkins user's plugin directory
     def self.plugins_dir
-      return File.join(self.home_dir, 'plugins')
+      if Facter.value(:osfamily) == 'OpenBSD'
+        return File.join(self.home_dir, '.jenkins', 'plugins')
+      else
+        return File.join(self.home_dir, 'plugins')
+      end
     end
   end
 end
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -41,6 +41,10 @@
 # service_ensure = 'running' (default)
 #   Status of the jenkins service.  running, stopped
 #
+# service_flags = undef (default)
+#   Service parameters for the jenkins service on OpenBSD, string, see daemon_flags
+#   in /etc/rc.d/jenkins for an example.
+#
 # config_hash = undef (Default)
 #   Hash with config options to set in sysconfig/jenkins defaults/jenkins
 #
@@ -53,6 +57,14 @@
 #
 # executors = undef (Default)
 #   Integer number of executors on the Jenkin's master.
+#
+# jarbin = 'jar' (Default system dependent)
+#   short or full path to the Java 'jar' binary. Full path necessary for systems
+#   where it is not in the default PATH, otherwise short path.
+#
+# javabin = 'java' (Default system dependent)
+#   short or full path to the Java 'java' binary. Full path necessary for systems
+#   where it is not in the default PATH, otherwise short path.
 #
 # slaveagentport = undef (Default)
 #   Integer number of portnumber for the slave agent.
@@ -163,11 +175,6 @@
 #   - Accepts input as array only.
 #   - Only effective if "proxy_host" and "proxy_port" are set.
 #
-# user = 'jenkins' (default)
-#
-# group = 'jenkins' (default)
-#
-#
 class jenkins(
   $version            = $jenkins::params::version,
   $lts                = $jenkins::params::lts,
@@ -179,6 +186,7 @@ class jenkins(
   $service_enable     = $jenkins::params::service_enable,
   $service_ensure     = $jenkins::params::service_ensure,
   $service_provider   = $jenkins::params::service_provider,
+  $service_flags      = undef,
   $config_hash        = {},
   $plugin_hash        = {},
   $job_hash           = {},
@@ -198,6 +206,8 @@ class jenkins(
   $manage_datadirs    = $jenkins::params::manage_datadirs,
   $localstatedir      = $::jenkins::params::localstatedir,
   $executors          = undef,
+  $jarbin             = $::jenkins::params::jarbin,
+  $javabin            = $::jenkins::params::javabin,
   $slaveagentport     = undef,
   $manage_user        = $::jenkins::params::manage_user,
   $user               = $::jenkins::params::user,
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -19,21 +19,28 @@ class jenkins::params {
   $package_name          = 'jenkins'
 
   $manage_datadirs = true
-  $localstatedir   = '/var/lib/jenkins'
 
   $manage_user  = true
-  $user         = 'jenkins'
   $manage_group = true
-  $group        = 'jenkins'
 
   case $::osfamily {
     'Debian': {
+      $user             = 'jenkins'
+      $group            = 'jenkins'
+      $jarbin           = 'jar'
+      $javabin          = 'java'
       $libdir           = '/usr/share/jenkins'
+      $localstatedir    = '/var/lib/jenkins'
       $package_provider = 'dpkg'
       $service_provider = undef
     }
     'RedHat': {
+      $user             = 'jenkins'
+      $group            = 'jenkins'
+      $jarbin           = 'jar'
+      $javabin          = 'java'
       $libdir           = '/usr/lib/jenkins'
+      $localstatedir    = '/var/lib/jenkins'
       $package_provider = 'rpm'
       case $::operatingsystem {
         'Fedora': {
@@ -51,8 +58,22 @@ class jenkins::params {
         }
       }
     }
+    'OpenBSD': {
+      $user             = '_jenkins'
+      $group            = '_jenkins'
+      $jarbin           = '/usr/local/jdk-1.8.0/bin/jar'
+      $javabin          = '/usr/local/jdk-1.8.0/bin/java'
+      $libdir           = '/usr/local/share/jenkins'
+      $localstatedir    = '/var/jenkins'
+      $package_provider = 'openbsd'
+    }
     default: {
+      $user             = 'jenkins'
+      $group            = 'jenkins'
+      $jarbin           = 'jar'
+      $javabin          = 'java'
       $libdir           = '/usr/lib/jenkins'
+      $localstatedir    = '/var/lib/jenkins'
       $package_provider = undef
       $service_provider = undef
     }
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -64,7 +64,7 @@ class jenkins::params {
       $jarbin           = '/usr/local/jdk-1.8.0/bin/jar'
       $javabin          = '/usr/local/jdk-1.8.0/bin/java'
       $libdir           = '/usr/local/share/jenkins'
-      $localstatedir    = '/var/jenkins'
+      $localstatedir    = '/var/jenkins/.jenkins'
       $package_provider = 'openbsd'
     }
     default: {
--- a/manifests/service.pp
+++ b/manifests/service.pp
@@ -6,12 +6,26 @@ class jenkins::service {
     fail("Use of private class ${name} by ${caller_module_name}")
   }
 
-  service { 'jenkins':
-    ensure     => $jenkins::service_ensure,
-    enable     => $jenkins::service_enable,
-    provider   => $jenkins::service_provider,
-    hasstatus  => true,
-    hasrestart => true,
+  case $::osfamily  {
+    'OpenBSD': {
+        service { 'jenkins':
+          ensure     => $jenkins::service_ensure,
+          enable     => $jenkins::service_enable,
+          provider   => $jenkins::service_provider,
+          flags      => $jenkins::service_flags,
+          hasstatus  => true,
+          hasrestart => true,
+        }
+    }
+    default: {
+      service { 'jenkins':
+        ensure     => $jenkins::service_ensure,
+        enable     => $jenkins::service_enable,
+        provider   => $jenkins::service_provider,
+        hasstatus  => true,
+        hasrestart => true,
+      }
+    }
   }
 
 }
--- a/lib/puppet/jenkins/facts.rb
+++ b/lib/puppet/jenkins/facts.rb
@@ -13,7 +13,7 @@ module Puppet
       # @return [NilClass]
       def self.install
         Facter.add(:jenkins_plugins) do
-          confine :kernel => [ "Linux", "OpenBSD" ]
+          confine :kernel => [ 'Linux', 'OpenBSD' ]
           setcode do
             Puppet::Jenkins::Facts.plugins_str
           end
--- a/lib/puppet/jenkins/facts.rb
+++ b/lib/puppet/jenkins/facts.rb
@@ -13,7 +13,7 @@ module Puppet
       # @return [NilClass]
       def self.install
         Facter.add(:jenkins_plugins) do
-          confine :kernel => 'Linux'
+          confine :kernel => [ "Linux", "OpenBSD" ]
           setcode do
             Puppet::Jenkins::Facts.plugins_str
           end
--- a/manifests/cli.pp
+++ b/manifests/cli.pp
@@ -23,7 +23,7 @@ class jenkins::cli {
       Anchor['jenkins::end']
 
   $jar = "${jenkins::libdir}/jenkins-cli.jar"
-  $extract_jar = "$::jenkins::jarbin -xf ${jenkins::libdir}/jenkins.war WEB-INF/jenkins-cli.jar"
+  $extract_jar = "${::jenkins::jarbin} -xf ${jenkins::libdir}/jenkins.war WEB-INF/jenkins-cli.jar"
   $move_jar = "mv WEB-INF/jenkins-cli.jar ${jar}"
   $remove_dir = 'rm -rf WEB-INF'
 
--- a/manifests/cli.pp
+++ b/manifests/cli.pp
@@ -23,7 +23,7 @@ class jenkins::cli {
       Anchor['jenkins::end']
 
   $jar = "${jenkins::libdir}/jenkins-cli.jar"
-  $extract_jar = "jar -xf ${jenkins::libdir}/jenkins.war WEB-INF/jenkins-cli.jar"
+  $extract_jar = "$::jenkins::jarbin -xf ${jenkins::libdir}/jenkins.war WEB-INF/jenkins-cli.jar"
   $move_jar = "mv WEB-INF/jenkins-cli.jar ${jar}"
   $remove_dir = 'rm -rf WEB-INF'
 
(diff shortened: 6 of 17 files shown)


manifests/params.pp
```diff
@@ -18,15 +17,43 @@
 
   case $::osfamily {
     'Debian': {
+      $jenkins_user     = 'jenkins'
+      $jenkins_group    = 'jenkins'
+      $repo             = true
+      $homedir          = '/var/lib/jenkins'
+      $jenkinsdir       = $homedir
+      $default_plugin_dir = "${jenkinsdir}/plugins"
```
jhoblitt: Might be worth lining up the indenting even though the linter won't complain about it.

manifests/plugin.pp
```diff
@@ -22,16 +22,24 @@
...
+  $jenkinsdir = $::jenkins::params::jenkinsdir
+  $username   = $::jenkins::params::jenkins_user
+  $group      = $::jenkins::params::jenkins_group
+
+  if $plugin_dir == '/var/lib/jenkins/plugins' {
+    $real_plugin_dir = $::jenkins::params::default_plugin_dir
+  } else {
+    $real_plugin_dir = $plugin_dir
+  }
+
```
jhoblitt: I'm not sure I follow the logic of what's happening here?

A selector would be more idiomatic way of assigning to $real_plugin_dir.

manifests/service.pp
```diff
@@ -6,11 +6,22 @@
     fail("Use of private class ${name} by ${caller_module_name}")
   }
 
-  service { 'jenkins':
-    ensure     => $jenkins::service_ensure,
-    enable     => $jenkins::service_enable,
-    hasstatus  => true,
-    hasrestart => true,
+  if $::osfamily != 'OpenBSD' {
```
jhoblitt: How about a case statement with a default?

manifests/slave.pp
```diff
@@ -150,7 +150,7 @@
     ensure => 'file',
     mode   => '0755',
     owner  => 'root',
-    group  => 'root',
+    group  => '0',
```
jhoblitt: I have a vague memory of a similar change in another module causing problems but I don't recall which module.  I'll see if I can find a reference.

buzzdeee: I've seen problems with the future parser, when the UID or GID is given literally and interpreted as integer, i.e.:
group => 0,

that usually caused trouble to me, since I run the future parser since a long time. However, '0' was the "safe" way to fix it ;)

manifests/plugin.pp
buzzdeee: changed to the suggested selector.
The idea behind is that, the default value for the parameter plugin_dir was originally hard-coded to '/var/lib/jenkins/plugins', and I did not wanted to break existing configurations that relied on it.
Since this is a defined type, and I remember I had trouble with just assigning 
$plugin_dir = $::jenkins::params::default_plugin_dir as the parameter to the defined type, even if the include ::jenkins::params is there in the list of parameters to the defined type. Therefore this kind of workaround.

buzzdeee:
I think I addressed all of your comments, Travis is still green, and it still seems to work for me ;)
let me know if there is more.
Commits squashed again.

buzzdeee:
rebased for your convenience

manifests/slave.pp
petems: Could you do something like this:

```
$rootgroup = $::osfamily ? {
    /(Free|Open)BSD/ => 'wheel',
    default          => 'root',
  }
```

matthewbarr: Probably needs to be in the params.pp, which is the right place for this kind of thing.

manifests/service.pp
```diff
@@ -6,11 +6,23 @@
...
-    ensure     => $jenkins::service_ensure,
-    enable     => $jenkins::service_enable,
-    hasstatus  => true,
-    hasrestart => true,
+  case $::osfamily  {
+    'OpenBSD': {
+        service { 'jenkins':
+          ensure     => $jenkins::service_ensure,
+          enable     => $jenkins::service_enable,
+          flags      => $jenkins::service_flags,
```
matthewbarr: Why does this need to be it's own case?  If flags is empty, it'll be fine, iirc.

manifests/slave.pp
buzzdeee: will look into that.

petems:
@buzzdeee There's upcoming OpenBSD support in Beaker (https://github.com/puppetlabs/beaker/pull/812) so there might be the option to add a beaker spec for this when it gets merged! :+1:

manifests/service.pp
buzzdeee: I had run into trouble with a different module I made aware of those flags. Short after merge upstream, someone who ran an older version of PE complained. That version he ran, I don't remember, wasn't aware of the flags parameterr to the service, and bailed out. Therefore the two cases here, If you don't care, I can have it without the case statement ;)

manifests/init.pp
```diff
@@ -41,12 +41,17 @@
 # service_ensure = 'running' (default)
 #   Status of the jenkins service.  running, stopped
 #
+# service_flags = undef (default)
+#   Service parameters for the jenkins service on OpenBSD, string, see daemon_flags
+#   in /etc/rc.d/jenkins for an example.
```
jhoblitt: Is there supposed to be a template for `/etc/rc.d/jenkins`?

lib/puppet/jenkins.rb
```diff
@@ -4,7 +4,11 @@ module Jenkins
     # @return [String] Full path to the Jenkins user's home directory
     def self.home_dir
       begin
-        return File.expand_path('~jenkins')
+        if Facter.value(:osfamily) == 'OpenBSD'
+          return File.expand_path('~_jenkins')
```
jhoblitt: is `_${USER}` an openbsd convention?

lib/puppet/jenkins.rb
```diff
@@ -13,7 +17,11 @@ def self.home_dir
 
     # @return [String] Full path to the Jenkins user's plugin directory
     def self.plugins_dir
-      return File.join(self.home_dir, 'plugins')
+      if Facter.value(:osfamily) == 'OpenBSD'
+        return File.join(self.home_dir, '.jenkins', 'plugins')
```
jhoblitt: Why does the path relative to the user's homedir need to be different on openbsd?

manifests/init.pp
```diff
@@ -147,6 +152,7 @@
   $package_provider   = $jenkins::params::package_provider,
   $service_enable     = $jenkins::params::service_enable,
   $service_ensure     = $jenkins::params::service_ensure,
+  $service_flags      = undef,
```
jhoblitt: If this is required, does it need a default value for openbsd in `jenkins::params`?

manifests/params.pp
```diff
@@ -18,16 +17,44 @@
...
+      $libdir             = '/usr/lib/jenkins'
+      $package_provider   = 'rpm'
+    }
+    'OpenBSD': {
+      $jenkins_user       = '_jenkins'
+      $jenkins_group      = '_jenkins'
+      $repo               = false
+      $homedir            = '/var/jenkins'
+      $jenkinsdir         = "${homedir}/.jenkins"
```
jhoblitt: Repeating my earlier comment, why doesn't `$homedir == $jenkinsdir` as on all the other supported platforms?

manifests/plugin.pp
```diff
@@ -22,16 +22,23 @@
   $config_content  = undef,
   $update_url      = undef,
   $plugin_dir      = '/var/lib/jenkins/plugins',
-  $username        = 'jenkins',
-  $group           = 'jenkins',
   $enabled         = true,
```
jhoblitt: This would be an API breaking change.  I think the parameters need to be retained.

manifests/plugin.pp
```diff
@@ -133,11 +140,11 @@
...
     exec { "create-pinnedfile-${name}" :
-      command => "touch ${plugin_dir}/${name}.jpi.pinned",
-      cwd     => $plugin_dir,
-      require => File[$plugin_dir],
-      path    => ['/usr/bin', '/usr/sbin', '/bin'],
-      onlyif  => "test -f ${plugin_dir}/${name}.jpi -a ! -f ${plugin_dir}/${name}.jpi.pinned",
+      command => "touch ${real_plugin_dir}/${name}.jpi.pinned",
+      cwd     => $real_plugin_dir,
+      require => File[$real_plugin_dir],
+      path    => ['/usr/bin', '/usr/sbin', '/bin', '/usr/local/bin'],
```
jhoblitt: `touch` lives under /usr/local/bin on openbsd?

manifests/plugin.pp
```diff
@@ -150,13 +157,13 @@
 
     exec { "download-${name}" :
       command => "rm -rf ${name} ${name}.hpi ${name}.jpi && wget --no-check-certificate ${download_url}",
-      cwd     => $plugin_dir,
```
jhoblitt: This needs to be rebased on master as https://github.com/jenkinsci/puppet-jenkins/pull/335 removed usage of wget from `plugin.pp`.

manifests/plugin.pp
```diff
@@ -22,16 +22,23 @@
...
 ) {
   include ::jenkins::params
 
-  $plugin            = "${name}.hpi"
-  $plugin_parent_dir = inline_template('<%= @plugin_dir.split(\'/\')[0..-2].join(\'/\') %>')
+  $plugin     = "${name}.hpi"
+  $homedir    = $::jenkins::params::homedir
+  $jenkinsdir = $::jenkins::params::jenkinsdir
+  $username   = $::jenkins::params::jenkins_user
+  $group      = $::jenkins::params::jenkins_group
```
jhoblitt: Setting the username and group in `jenkins::params` seems unrelated to adding openbsd support?

manifests/plugin.pp
```diff
@@ -22,16 +22,23 @@
...
+
+  $real_plugin_dir = $plugin_dir ? {
+    '/var/lib/jenkins/plugins' => $::jenkins::params::default_plugin_dir,
+    default                    => $plugin_dir,
+  }
```
jhoblitt: The default value being assign is not the newly created variable named `default_plugin_dir` -- the variable naming is difficult to understand.  At a minimum, defined type's parameter default for $plugin_dir should be changed to `undef` and I think $default_plugin_dir should be renamed (if it is needed at all).

manifests/init.pp
buzzdeee: nope, OpenBSD jenkins package comes with a well working rc script, see: http://cvsweb.openbsd.org/cgi-bin/cvsweb/~checkout~/ports/devel/jenkins/pkg/jenkins.rc?rev=1.6&content-type=text/plain

manifests/init.pp
buzzdeee: the defaults given in the OS in the rc script are fine, no need for it in the params.pp

manifests/params.pp
buzzdeee: I'm not the jenkins package maintainer. But as far as I can see, it's the way it is since ever on OpenBSD.
What is done is just using downloading the war file, deployed to /usr/local/share/jenkins. No JENKINS_HOME or the like ENV variable is set. As far as I can see, without such variable, the war file defaults to the ~/.jenkins directory of the jenkins user?

manifests/plugin.pp
buzzdeee: yes, will do

manifests/plugin.pp
buzzdeee: iirc, it's used a bit way down, and since they are differnet on OpenBSD than elsewhere, I did it that way.
Have to review, since after looking at it again, I think now, there might be dragons lurking anyways ;)

manifests/plugin.pp
buzzdeee: I'll see to rework that.

manifests/plugin.pp
buzzdeee: good catch, nope it doesn't

manifests/plugin.pp
buzzdeee: alright

lib/puppet/jenkins.rb
buzzdeee: yes, more or less every port that has a daemon that not necessarily must run as root, has such users/group names, see: http://cvsweb.openbsd.org/cgi-bin/cvsweb/~checkout~/ports/infrastructure/db/user.list?rev=1.249&content-type=text/plain

lib/puppet/jenkins.rb
buzzdeee: This is how the jenkins package works on OpenBSD, jenkins looks and picks up the plugins from ~_jenkins/.jenksins/plugins

petems:
@buzzdeee Got a working OpenBSD Beaker test :dancer: :

```ruby
require 'spec_helper_acceptance'

describe 'jenkins class on OpenBSD', :if => fact('osfamily') == 'OpenBSD' do

  before(:all) do
    shell('rm -rf /etc/puppet/modules/java')
    shell('pkg_add -I git')
    shell('git clone https://github.com/puppetlabs/puppetlabs-java /etc/puppet/modules/java')

    pkg_conf_pp = <<-EOS

    $source = "http://ftp3.usa.openbsd.org/pub/OpenBSD/${::kernelmajversion}/packages/${::architecture}/"

    file { '/etc/pkg.conf':
      ensure  => file,
      owner   => 'root',
      group   => 'wheel',
      mode    => '0644',
      content => "# OpenBSD pkg.conf\ninstallpath=${source}\n",
    }
  EOS

    apply_manifest(pkg_conf_pp, :catch_failures => true)
  end

  context 'default parameters' do
    it 'should work with no errors' do
      pp = <<-EOS
      class {'jenkins':}
      EOS

      # Run it twice and test for idempotency
      apply_manifest(pp, :catch_failures => true)
      apply_manifest(pp, :catch_changes => true)
    end

    describe port(8000) do
      it {
        sleep(10) # Jenkins takes a while to start up
        should be_listening
      }
    end

    describe service('jenkins') do
      it { should be_running }
    end

  end
end
```

buzzdeee:
@petems, for the $source in the beaker test you might want to use:
"http://ftp.openbsd.org/pub/OpenBSD/${::kernelmajversion}/packages/${::hardwareisa}/"

so ::hardwareisa instead of ::architecture, There are architectures/platforms, that differ, i.e. macppc/powerpc, and a few others.
Anyways, since Jenkins is java based, and java is only for i386 and amd64, both are safe ;)

Do you want me to integrate that? Haven't used beaker before, so it would not incredibly easy to test for me ;)

cheers,

petems:
@buzzdeee I can pull it in as a separate PR :+1:

petems:
See https://github.com/jenkinsci/puppet-jenkins/pull/342

buzzdeee:
that's fine with me, thanks!
BTW, that failed travis-ci build, just needs to be restarted. I also got some odd failings on my repo when I updated and pushed changes, usually a restart helped to do tests in different random order, and all went green.

I kept the large number of changes for the time being, for the case you want to see better what I did. I can squash/rebase whenever necessary or wanted.

manifests/plugin.pp
```diff
@@ -32,8 +32,22 @@
...
+    default => $username,
+  }
+  $jgroup = $group ? {
+    ''      => $::jenkins::params::jenkins_group,
+    default => $group,
+  }
+  $jplugin_dir = $plugin_dir ? {
+    ''      => $::jenkins::params::plugin_dir,
+    default => $plugin_dir,
+  }
```
buzzdeee: I had to do above dance with the parameters,
I just tried to pull in $plugin_dir = $::jenkins::params::plugin_dir
and also the others, in the parameter list of the defined type, but then all the time, the tests failed. It seemed that Puppet didn't pulled that info in at that time.

I had to set the default values for the parameters to '', because the selector didn't wanted to work when they were set to undef. 

Alternatively, I could set these variables default to undef and do check like:

if $plugin_dir {
  $jplugin_dir = $plugin_dir
} else {
  $jplugin_dir = $::jenkins::params::plugin_dir
}

if that is preferred, or let me know if there is a better way to work around that.

jhoblitt: A `:pre_condition` block that pulls in `jenkins/jenkins::params`would likely be needed for those tests to work.  However, that would introduce evaluation order dependent behavior for users.  My opinion is that introducing parse order requirements at this point would be a major API break.

Unlike the `inherits` hack that works for classes, there's no way [that I'm aware of] internal to a defined type to require that a class be evaluated before it in the manifest.  It can be done external to the define by placing constraints using resource collectors (which have side effects on virtual resources) or if you know the name of the define.  I used the later approach in my selenium module.  Eg., https://github.com/jhoblitt/puppet-selenium/blob/master/manifests/hub.pp#L14-L23 Which isn't exposed to the user as its wrapped up in a class.  I think the if/else approach you've described or a custom function is the most robust approach.

jhoblitt:
@buzzdeee I'd like to suggest that you try to split all of the non-specific openbsd values out of this branch and create a new PR for it.  I think there's some valuable house keeping in this PR, in particular having a canonical value for the jenkins user/group, that should be fairly straight forward to merge.  That should ultimately make it easier to keep this PR rebased on the current master.

buzzdeee:
I'll see what I can do to split off the PR into multiple, just need to find the time, hopefully on the weekend, or next week.

manifests/plugin.pp
buzzdeee: I figured that I can do something like:

define jenkins::plugin (
foo,
) {
Class['jenkins::params'] ->
Jenkins::Plugin[$name]
}

to ensure that the jenkins::params are evaluated before the define.

See PR #356 , first start of to configure the user/group in params.pp

jhoblitt:
@buzzdeee I think if you rebase upon the current master this will be a much smaller PR now.

buzzdeee:
yay, will hopefully find time this week

cheers,

olevole:
Any update for this Issue ? I'm interesting in OpenBSD and FreeBSD support

bastelfreak:
Hi people. This PR is pretty huge and got quite some feedback. But since it got stalled, I'm going to close it. If anybody is still interested in it, please reopen/resubmit it and rebase.

PR #306, opened 2015-05-21
Set no proxy for plugins

ricfeatherstone:
When downloading plugins proxy is set for http and https but no proxy is not.

This causes issues when wanting to configure the Jenkins proxy details but host plugins internally.

### DIFF
--- a/manifests/plugin.pp
+++ b/manifests/plugin.pp
@@ -122,10 +122,14 @@ define jenkins::plugin(
 
   if (empty(grep([ $::jenkins_plugins ], $search))) {
     if ($jenkins::proxy_host) {
+      unless empty($jenkins::no_proxy_list) {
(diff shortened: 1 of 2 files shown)


jhoblitt:
This looks reasonable.

The unit tests for `jenkins::plugin` are really lacking.  Would you be willing to add coverage for the two exec resources this change will effect?

I'd also suggest squashing your history to remove the merge commits.

ricfeatherstone:
OK, history squashed and existing tests updated.

jhoblitt:
@ricfeatherstone I really do appreciate your work on this PR!  Since it was opened, all of exec resources have been removed from from `jenkins::plugin`.  Additionally, the `archive::download` resource respects `$jenkins::proxy_host`.  Regretfully, I'm going to close this PR but please do open a new issue/PR if there are still difficulties with proxy support.

PR #311, opened 2015-05-26
added jenkins::credentials class

ricoli:
renamed the defined type to be jenkins:credential

### DIFF
--- a/README.md
+++ b/README.md
@@ -290,7 +290,7 @@ Alternatively, using hiera:
 jenkins::credential_hash:
   github-deploy-key:
     password: ''
-    private_key_or_path => |
+    private_key_or_path: |
       AVERYLONGPRIVATEKEYHERE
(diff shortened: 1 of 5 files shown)


jhoblitt:
This seems reasonable, and does make the class naming more consistent, but it is an API breaking change as existing usage of `jenkins::credentials` will fail.  I think this either needs to be a major version bump (which I would be OK with) or reworked so the existing API `jenkins::credentials` semantics are preserved.

ricoli:
Indeed it does break compatibility, so would be great to know if a major bump is OK. Otherwise I could rename the class to something not as nice, like credentials_hash or something like that

jhoblitt:
After having spent several days digging through the credentials plugin source, I think it's probably better to leave the defined type with the pluralized name.  The plural form appears to be used consistently in the API across credential(s) related plugins.

PR #312, opened 2015-05-27
Added -i option in class jenkins::cli::reload when security is enabled

rmestrum:
When security is enabled the class jenkins::cli::reload will fail because there is no SSH private key file passed that's needed for authentication.


jhoblitt:
I believe this is a duplicate of #305 .

jhoblitt:
This should be resolved by #352

PR #319, opened 2015-06-09
Adding optional description to slave

tstibbs:
Previously if you created a slave using this manifest, it would show up in the master as something like `Slave MyNewSlave (Swarm slave from 192.168.0.10 : null)`. This pull request allows you to specify a description to put in place of the 'null'.


rtyler:
I do wish there were automated tests for this, but I don't have a good suggestion for how to write those in a non-insane fashion :)

jhoblitt:
@rtyler unit tests can inspect the interpolated template and  acceptance tests could look at the installed init script or the process title.

PR #322, opened 2015-06-16
Systemd

timemaster5:
This commit adds support for systemd based distros and upgrades swarm to 1.24 version - is there any reason to stay with older one ? Tested on Fedora21 and Centos7.1.

### DIFF
--- a/manifests/slave.pp
+++ b/manifests/slave.pp
@@ -141,6 +141,14 @@ class jenkins::slave (
         before => Service['jenkins-slave'],
       }
     }
+    'RedHat': {
+      file { '/etc/systemd/system/jenkins-slave.service':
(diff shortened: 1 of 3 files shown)


manifests/slave.pp
```diff
@@ -141,6 +141,14 @@
         before => Service['jenkins-slave'],
       }
     }
+    'RedHat': {
```
pkittenis: Not all redhat releases run systemd..

jhoblitt: Indeed.  EL < 7 does not and there are likely amazon linux variants still in use that are pre-systemd as well.

jhoblitt:
Systemd support for swarm clients was merged in #691 / #692.

PR #328, opened 2015-06-24
Support configuring a yum proxy server

tosmi:
This pull request adds a new parameter yum_proxy to init.pp. It's used in repo/el.pp to configure the jenkins yum repository with a proxy server. 

This allows installing jenkins via yum behind a proxy server.

### DIFF
--- a/manifests/repo/el.pp
+++ b/manifests/repo/el.pp
@@ -7,6 +7,8 @@ class jenkins::repo::el
     fail("Use of private class ${name} by ${caller_module_name}")
   }
 
+  $yum_proxy = $::jenkins::yum_proxy
+
   if $::jenkins::lts  {
     yumrepo {'jenkins':
       descr    => 'Jenkins',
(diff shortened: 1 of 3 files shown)


manifests/repo/el.pp
```diff
@@ -26,4 +28,11 @@
       enabled  => 1,
     }
   }
+
+  if $yum_proxy {
+    Yumrepo['jenkins'] {
+      proxy => $yum_proxy,
+    }
```
jhoblitt: Does this need to be done as a resource default?  I feel that they are somewhat of an anti-pattern until if/when the scope is restricted.  I've battled endless problems with defaults leaking out and causing unintended side-effects.

tosmi: not necessarily. this was the first solution that came to my mind without thinking to much about it...

i could move the proxy parameter directly to the yumrepo resource declarations. should we keep 'undef' for the default case or should we convert the parameter to 'absent' in el.pp?

i was also thinking about the name 'yum_proxy'. maybe we should call it 'repo_proxy' as this could also be used for apt and zypper repos. 

thanks for you feedback

jhoblitt: `repo_proxy` might be a bit better long term since fedora 22 has replaced yum with dnf.

PR #333, opened 2015-07-08
Fix jenkins::plugin with create_user false

valtri:
We have setup with jenkins running under tomcat (user tomcat7), but still it is good idea to have Jenkins files in /var/lib/jenkins.

This patch fixes plugin installation in this use-case - with external user already created and also when it is different from /var/lib/jenkins owner.


rtyler:
This is failing in travis because the `rake lint` task isn't passing.

valtri:
I'm sorry about that! It seems after fixing the whitspaces, 'bundle exec rake' is OK. I'll open another pull request.

PR #339, opened 2015-07-13
Allow to setup LDAP auth

mescanef:
Be able to setup LDAP auth on Jenkins master. This resulted in splitting
set security method into three methods for setting:

ldap auth
password auth
unsecured
Also give choice to turn on the slave to master feature during
deployment.


manifests/security.pp
```diff
@@ -17,17 +17,91 @@
...
 class jenkins::security (
-  $security_model = undef,
+  $jenkins_management_email         = '',
+  $jenkins_management_login         = '',
+  $jenkins_management_name          = '',
+  $jenkins_management_password      = '',
+  $jenkins_s2m_acl                  = false,
+  $jenkins_ssh_private_key_contents = '',
+  $jenkins_ssh_public_key_contents  = '',
+  $ldap_access_group                = '',
```
rtyler: This is a huge amount of parameters for the jenkins::security class. I suggest changing this to take a hash called `ldap_config`. You could even set the defaults in a container class like `jenkins::ldap::config` or something like that.

manifests/security.pp
```diff
@@ -17,17 +17,91 @@
...
+  $ldap_user_search_base            = '',
+  $security_model                   = undef,
 ){
   validate_string($security_model)
 
   include ::jenkins::cli_helper
 
+  $libdir = $::jenkins::libdir
+
+  file { "${libdir}/.ssh/id_rsa" :
```
rtyler: The management of some Jenkins public/private SSH keys seems unrelated to this LDAP configuration change, what's it here for?

manifests/security.pp
```diff
@@ -17,17 +17,91 @@
...
+    content => "${jenkins_ssh_public_key_contents} jenkins@${::fqdn}",
+    replace => true,
+    require => File["${libdir}/.ssh"],
+  }
+
+  if ($security_model == 'unsecured') {
+    $security_opt_params = 'set_security_unsecured'
+  }
+
+  if ($security_model == 'ldap') {
```
rtyler: One suggestion, since this is a lot to place on the command line. What about writing a configuration file to disk temporarily and then having the groovy helper read that configuration file?

I worry that there command line might end up being too long

jhoblitt: In my native types and provider branch I've been passing json into the puppet_helper on the stdin, which has worked well.

jhoblitt: Here's a specific example of handling json input: https://github.com/jhoblitt/puppet-jenkins/blob/feature/types_and_providers/files/puppet_helper.groovy#L209-L212

jhoblitt:
@mescanef My work to support configuring the security realm and authentication strategy via native types was merged in #382.  Do you think this work could be merge with the groovy `get_security_realm()` / `get_authorization_strategy()` methods on master and the providers for `jenkins_security_realm` and `jenkins_authorization_strategy`?  @daenney is also interested in configuring ldap.

rhoml:
Any news on this PR?

jhoblitt:
@rhoml At a minimum it needs to be rebased on the current master and I think it should be converted to passing json instead of CLI args.  I think it would be less work to add this to the `jenkins_security_realm/jenkins_authorization_strategy` types.

manifests/security.pp
matonb: Nothing like reviving old branches...

@jhoblitt Your example json URL appears to be dead, are you still using this branch to setup Jenkins LDAP authentication ?

jhoblitt: @matonb That code has been merged into this module.

bastelfreak:
Hello everybody. I'm going to close this PR due to inactivity. Please reopen and rebase the PR if anybody is still interested in it.

PR #340, opened 2015-07-14
(RFC - not ready for merge) use camptocamp/archive for all downloads

jhoblitt:
Replaces direct usage of wget from exec and nanliu/staging.  Intended to be an
interim solution until nanliu/archive gains proxy support.

### DIFF
--- a/manifests/direct_download.pp
+++ b/manifests/direct_download.pp
@@ -25,11 +25,24 @@ class jenkins::direct_download {
   validate_absolute_path($local_file)
 
   if $::jenkins::version != 'absent' {
+    $enable_checksum = $::jenkins::direct_download_checksum ? {
+      undef   => false,
+      default => true,
+    }
+
     # make download optional if we are removing...
-    staging::file { $package_file:
-      source => $jenkins::direct_download,
-      target => $local_file,
-      before => Package[$::jenkins::package_name],
+    archive::download { $package_file:
+      url              => $::jenkins::direct_download,
+      src_target       => $::jenkins::package_cache_dir,
+      allow_insecure   => true,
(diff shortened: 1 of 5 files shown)


jhoblitt:
This PR is in response to discussion in PR #336 about converging on a single method of downloading files.  I've considered a few different modules and the tl;dr is that I'm proposing to exclusively use `camptocamp/archive` until `nanliu/archive` gains proxy support.

I try to use supported/approved modules when possible on the theory that they are less likely to cause an environment conflict.  AFAIK, there are two blessed, popular, and API stable download modules for posix like systems.  `nanliu/staging` & ` maestrodev/wget`.  The former does not have proxy or checksum support.  The later has both but proxy support is only via facts/top level variables. Neither appear to be an ideal fit.

I suspect that `nanliu/archive`, being a native type implementation, is "the future" for downloading files. However, it currently lacks proxy support and presently requires a fair number of gems to be installed on the agent . `camptocamp/archive` has all of the features this module desires but I dislike that the destination filename is implicit based on the resource name, that it creates a file resource for the downloaded file only upon `ensure => absent`, and that the module name conflicts with `nanliu/archive`.

rtyler:
@jhoblitt in general I really like this change, the failing tests concern me though. I might take a whack at fixing it later today if I don't run out of steam first

jhoblitt:
@rtyler Ping me if you don't get to it and I'll try to take a stab and fixing up the tests tomorrow.

rtyler:
@jhoblitt I definitely didn't get to this :P

jhoblitt:
I think the rough consensus is to switch to puppet-community/puppet-archive, so I'm going to retract this PR.

PR #345, opened 2015-07-15
Parameterises the Java package name

petems:
On CentOS 5, the Puppetlabs Java module defaults to Java 1.6

However, this will break Jenkins:

```
/usr/bin/java -jar /usr/lib/jenkins/jenkins-cli.jar -s http://127.0.0.1:8080 groovy /usr/lib/jenkins/puppet_helper.groovy set_num_executors 42
Exception in thread "main" java.lang.UnsupportedClassVersionError: hudson/cli/CLI : Unsupported major.minor version 51.0
	at java.lang.ClassLoader.defineClass1(Native Method)
	at java.lang.ClassLoader.defineClass(ClassLoader.java:643)
	at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
	at java.net.URLClassLoader.defineClass(URLClassLoader.java:277)
	at java.net.URLClassLoader.access$000(URLClassLoader.java:73)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:212)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.net.URLClassLoader.findClass(URLClassLoader.java:205)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:323)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:294)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:268)
Could not find the main class: hudson.cli.CLI. Program will exit.
```

So you have to tell it to install java-1.7.0-openjdk.x86_64 for Java.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -110,6 +110,9 @@
 #   - use puppetlabs-java module to install the correct version of a JDK.
 #   - Jenkins requires a JRE
 #
+# java_package_name = undef (default)
+#   - choose the name of the package to install with puppetlabs-java if the if install_java parameter is true
+#   - defaults to undef, which will use the default taken from the puppetlabs-java module from the params
 #
 # cli = false (default)
 #   - force installation of the jenkins CLI jar to $libdir/cli/jenkins-cli.jar
@@ -155,6 +158,7 @@ class jenkins(
   $user_hash          = {},
   $configure_firewall = undef,
   $install_java       = $jenkins::params::install_java,
+  $java_package_name  = undef,
   $repo_proxy         = undef,
   $proxy_host         = undef,
   $proxy_port         = undef,
@@ -187,7 +191,8 @@ class jenkins(
 
   if $install_java {
     class {'java':
-      distribution => 'jdk'
+      distribution  => 'jdk',
+      name          => $java_package_name,
     }
   }
 
--- a/spec/classes/jenkins_spec.rb
+++ b/spec/classes/jenkins_spec.rb
@@ -33,6 +33,11 @@ describe 'jenkins', :type => :module do
       it { should_not contain_class 'jenkins::repo' }
     end
 
+    describe 'with java_package_name' do
+      let(:params) { { :java_package_name => 'java-1.7.0-openjdk.x86_64' } }
+      it { should contain_class('java').with({'name' => 'java-1.7.0-openjdk.x86_64'}) }
+    end
+
     describe 'with only proxy host' do
       let(:params) { { :proxy_host => '1.2.3.4' } }
       it { should_not contain_class('jenkins::proxy') }


manifests/init.pp
```diff
@@ -187,7 +191,8 @@
 
   if $install_java {
     class {'java':
-      distribution => 'jdk'
+      distribution          => 'jdk',
```
matthewbarr: That'a a lot of white space, btw.

manifests/init.pp
```diff
@@ -155,6 +158,7 @@
   $user_hash          = {},
   $configure_firewall = undef,
   $install_java       = $jenkins::params::install_java,
+  $java_package_name  = undef,
```
matthewbarr: Yiou aren't specifying any default.  Is a default needed?  If yes, then we should probably be dealing with it in params.pp, and pulling the setting in here., similar to the prior line.

manifests/init.pp
petems: Yeah, I'm not sure why I went so overboard here, should've just made the arrows match for linting. Let me fix.

petems: Fixed :+1:

manifests/init.pp
petems: Yeah, I was going to ask about this. The Jenkins install failed on 1.6, so we should probably default to 1.7. But I wasn't sure if there were any OS's that had to use 1.6.

If there's no issue, this could be made a param and inherited.

matthewbarr: You'd probably want to make it only default to 1.7 on cent/rhel 5.    most of the rest use default, and 1.7 is deprecated arlready.  newer ubuntu's use 1.8, for example.  

But that's what params is for, doing OS based defaults :)

petems: Yeah, there's defaults on the PL Java module, but some of those defaults aren't compatible with Jenkins. I think Jenkins requires at least 1.7 from the errors I was seeing.

jhoblitt:
In terms of general approach, I'd rather not try to pass a bunch of parameters to an external module.  I think its generally more flexible to allow the user to disable the automatic management so they can manage the module as they need.

petems:
@jhoblitt I can agree with that. A good middle group would be to add a RHEL5 example to the `examples/` folder with an example RHEL5 specific profile with the Java version defined and such. Sound good?

jhoblitt:
@petems That would be my preference over having to hardcode a package name based on the assumption that EL5 is fairly small part of the user base.  Others may disagree?

Thinking about it some more, an ideal solution would be for the java module to accept a version constraint, E.g. jdk >= 7, so that low level details don't need to be hard coded into consumer modules.  However, that would be quite a bit different [and a lot of work to implement] than the current [`version`](https://github.com/puppetlabs/puppetlabs-java/#version) param.

PR #346, opened 2015-07-16
Adds Examples for various platforms for Jenkins

petems:
Examples are moved from tests/ -> examples as per https://docs.puppetlabs.com/puppet/latest/reference/modules_fundamentals.html#module-layout


jhoblitt:
Merged as #390 with trivial linter fixes.

Issue #132, opened 2014-05-13
warning: Could not load fact file /etc/puppet/modules/jenkins/lib/facter/jenkins.rb: no such file to load -- puppet/jenkins/facts


leopoiroux:
Hi,
can you help me? 
I still have this message when I execute a "puppet apply". 

warning: Could not load fact file /etc/puppet/modules/jenkins/lib/facter/jenkins.rb: no such file to load -- puppet/jenkins/facts

I'm on :
- Debian 7.4 
- Puppet 2.7.23 
- Facter 1.6.10

PR #347, opened 2015-07-17
Fix require paths

joshuaspence:
This fixes the `require` paths on Ruby 1.8.7 with Puppet 2.7. See issue #132.


rtyler:
@joshuaspence I don't have 2.7 environments floating around anymore. I'm curious what level of testing has been applied to this change since there's nothing automated

joshuaspence:
We've been running with this modification for more than 6 months in a 2.7 environment.

PR #351, opened 2015-07-23
add darwin/osx support to slave class

adamcstephens:
This utilizes the existing defaults file, along with a wrapper script and launchd service. There is no wget by default on OSX so use curl there (could be used everywhere, or instead use a puppet module such as archive).

### DIFF
--- a/manifests/slave.pp
+++ b/manifests/slave.pp
@@ -126,15 +126,6 @@ class jenkins::slave (
     }
   }
 
-  exec { 'get_swarm_client':
-    command => "wget -O ${slave_home}/${client_jar} ${client_url}/${client_jar}",
-    path    => '/usr/bin:/usr/sbin:/bin:/usr/local/bin',
-    user    => $slave_user,
-  #refreshonly => true,
-    creates => "${slave_home}/${client_jar}",
-  ## needs to be fixed if you create another version..
-  }
-
 # customizations based on the OS family
   case $::osfamily {
     'Debian': {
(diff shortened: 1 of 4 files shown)


spec/classes/jenkins_slave_spec.rb
```diff
@@ -67,6 +68,19 @@
     end
   end
 
+#  describe 'Darwin' do
```
rtyler: Is there any reason for these examples to be commented out?

manifests/slave.pp
```diff
@@ -145,31 +114,110 @@
...
 
-  file { '/etc/init.d/jenkins-slave':
-    ensure => 'file',
-    mode   => '0755',
-    owner  => 'root',
-    group  => 'root',
-    source => "puppet:///modules/${module_name}/jenkins-slave.${::osfamily}",
-    notify => Service['jenkins-slave'],
+  case $::kernel {
+    'Linux': {
```
rtyler: This case statement smells a bit to me, would it be possible to refactor this into a `jenkins::slave::linux` and a `jenkins::slave::darwin` module which would then be included depending on the kernel?

rtyler:
@adamcstephens overall, looking very good! some minor comments that I've added; can't wait to get this merged

spec/classes/jenkins_slave_spec.rb
adamcstephens: I'm still learning rspec, and can't figure out how to get the tests to ignore java install on darwin.

```
Failures:

  1) jenkins::slave Darwin with slave_name behaves like using slave_name should contain File[/home/jenkins-slave/jenkins-slave] with content =~ /^CLIENT_NAME="jenkins-slave"$/
     Failure/Error: it { should contain_file(slave_runtime_file).with_content(/^CLIENT_NAME="jenkins-slave"$/) }
     Puppet::Error:
       Could not find resource 'Class[Java]' for relationship on 'Service[jenkins-slave]' on node myosx.local
     Shared Example Group: "using slave_name" called from ./spec/classes/jenkins_slave_spec.rb:80
     # ./spec/classes/jenkins_slave_spec.rb:43:in `block (3 levels) in <top (required)>'
```

rtyler:
I'm going to merge this and try to clean it up a bit myself before release

spec/classes/jenkins_slave_spec.rb
rtyler: so the puppetlabs/java module doesn't support Darwin so I think part of why this is failing.

adamcstephens:
Ok, sorry. I started on the refactor earlier in the week but work got the better of me.

rtyler:
@adamcstephens no big deal, this has since been released :)

PR #354, opened 2015-08-05
Add support for groups to slave user

jyaworski:
Add groups to the slave user.

### DIFF
--- a/manifests/slave.pp
+++ b/manifests/slave.pp
@@ -87,6 +87,7 @@ class jenkins::slave (
   $slave_uid                = undef,
   $slave_home               = '/home/jenkins-slave',
   $slave_mode               = 'normal',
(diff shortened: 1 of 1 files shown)


jhoblitt:
I would prefer that consumers set `manage_slave_user => false` and manage  the user resource directly rather than reproducing the entire user type's interface via this [or any class].  That approach is both more flexible and less maintenance burden -- others should sound in as to their opinion.

If this PR is to be merged, the new parameter should be validated and some test coverage is desirable.

jyaworski:
@jhoblitt I'm actually on your side on this one. I apparently overlooked that parameter, which fits my use-case.

I'll leave this open for further comment.

matthewbarr:
And I'd agree with all of you. There's a limit to flexibility...

PR #356, opened 2015-08-10
Make the jenkins system user and group OS dependent. 

buzzdeee:
Default values are set in and picked up from params.pp.

The two new parameters are:
  jenkins_user
  jenkins_group

The specs for the proxy class, and the plugin define added facts
to pin it to osfamily/operatingsystem RedHat, in order to let
them work in the future, when there may be other OS with different
users/groups out there.

I haven't made the jenkins_user/jenkins_group overridable as
parameter to init.pp, since I found that not all parameters defined
in params.pp are able to be overridden.

This is a follow up of PR #304 , splitting parts off the whole PR.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -18,14 +18,20 @@ class jenkins::params {
 
   case $::osfamily {
     'Debian': {
+      $jenkins_user       = 'jenkins'
+      $jenkins_group      = 'jenkins'
       $libdir           = '/usr/share/jenkins'
       $package_provider = 'dpkg'
     }
     'RedHat': {
+      $jenkins_user       = 'jenkins'
+      $jenkins_group      = 'jenkins'
       $libdir           = '/usr/lib/jenkins'
       $package_provider = 'rpm'
     }
     default: {
+      $jenkins_user       = 'jenkins'
+      $jenkins_group      = 'jenkins'
       $libdir = '/usr/lib/jenkins'
       $package_provider = false
     }
--- a/manifests/plugin.pp
+++ b/manifests/plugin.pp
@@ -22,8 +22,8 @@ define jenkins::plugin(
   $config_content  = undef,
   $update_url      = undef,
   $plugin_dir      = '/var/lib/jenkins/plugins',
-  $username        = 'jenkins',
-  $group           = 'jenkins',
+  $username        = '',
+  $group           = '',
   $enabled         = true,
   $create_user     = true,
   $source          = undef,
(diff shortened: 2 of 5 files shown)


rtyler:
@jhoblitt since you were following along more closely with #304, I trust that you'll review this? pretty plz :)

manifests/params.pp
```diff
@@ -18,14 +18,20 @@
 
   case $::osfamily {
     'Debian': {
+      $jenkins_user       = 'jenkins'
+      $jenkins_group      = 'jenkins'
```
jhoblitt: trivial quibble but the whitespace doesn't line up.

manifests/params.pp
```diff
@@ -18,14 +18,20 @@
...
       $libdir           = '/usr/share/jenkins'
       $package_provider = 'dpkg'
     }
     'RedHat': {
+      $jenkins_user       = 'jenkins'
+      $jenkins_group      = 'jenkins'
```
jhoblitt: whitespace

manifests/params.pp
```diff
@@ -18,14 +18,20 @@
...
       $libdir           = '/usr/lib/jenkins'
       $package_provider = 'rpm'
     }
     default: {
+      $jenkins_user       = 'jenkins'
+      $jenkins_group      = 'jenkins'
```
jhoblitt: whitespace (sort of) -- perhaps fixup the successive lines

manifests/params.pp
```diff
@@ -18,14 +18,20 @@
 
   case $::osfamily {
     'Debian': {
+      $jenkins_user       = 'jenkins'
```
jhoblitt: Since all the values for `$jenkins_user` and `$jenkins_group` are the same, they don't need to be in the case statement at all.

manifests/plugin.pp
```diff
@@ -22,8 +22,8 @@
   $config_content  = undef,
   $update_url      = undef,
   $plugin_dir      = '/var/lib/jenkins/plugins',
-  $username        = 'jenkins',
-  $group           = 'jenkins',
+  $username        = '',
+  $group           = '',
```
jhoblitt: I can't come up with a [good] reason why the plugins should be owned by a different user than the jenkins master. I feel we should probably depreciate / ignore these params and allow overriding the default user/group only via params on the `jenkins` class.

jhoblitt:
I'm going to take a stab at fixing up this PR.

PR #359, opened 2015-08-13
Adapt soon to be deprecated parameters (apt module)

fadenb:
Next major release of apt module will deprecate some parameters. They
were replaced by parameters that expect hashes.

### DIFF
--- a/manifests/repo/debian.pp
+++ b/manifests/repo/debian.pp
@@ -11,22 +11,30 @@ class jenkins::repo::debian
 
   if $::jenkins::lts  {
     apt::source { 'jenkins':
-      location    => 'http://pkg.jenkins-ci.org/debian-stable',
-      release     => 'binary/',
-      repos       => '',
-      key         => '150FDE3F7787E7D11EF4E12A9B7D32F2D50582E6',
-      key_source  => 'http://pkg.jenkins-ci.org/debian/jenkins-ci.org.key',
-      include_src => false,
+      location => 'http://pkg.jenkins-ci.org/debian-stable',
+      release  => 'binary/',
+      repos    => '',
+      key      => {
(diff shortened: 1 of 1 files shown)


madAndroid:
As far as I know, the 2.1 release of the apt module is backwardly compatible with 1.8 => https://github.com/puppetlabs/puppetlabs-apt/blob/master/CHANGELOG.md#2015-06-16---supported-release-210

Is this change then no longer required?

fadenb:
The current version is warning that it will be deprecated in the next major
version (I believe they mean 3.x.x).
I have seen no indication that they no longer plan to do it.

madAndroid:
Ah, okay .. I've just had issues with dependencies with other modules that have specified puppetlabs-apt `>= 2.0` - and was hoping that's not going to be the case with this module as well. Having had another look, looks like that was dealt with in 9846c6a21d9a7be13100fc9afed0d321b8696a7d, so it's all good :)

rtyler:
@fadenb I'm not following the development of the apt module and our current version range in `metadata.json` is: `>= 0.0.3 < 3.0.0`. 

Is this change still compatible with such early versions of puppetlabs/apt or will we need to modify our dependency requirement?

fadenb:
I will test compatibility with the older versions and let you know as soon as I am done.

rtyler:
@fadenb sounds good, thanks!

rtyler:
*bump*

daenney:
This change should only happen if you specify that you need puppetlabs-apt `>=2.0.0`, it will not work with puppetlabs-apt 1.x.

The parameters will remain in place until we release puppetlabs-apt 3.x, at which point the old way (without the hashes) will throw a compilation error.

rgevaert:
Now that version 4 of the apt module has been released the $include_src and  $key_source params are gone.

mmoll:
this got in with 82de1b2 - this PR can get closed

Issue #366, opened 2015-09-04
tool_locations doesn't work


patcadelina:
When defining a tool location as:

    class { 'jenkins::slave':
        slave_name     => 'slave_1',
        ...
        tool_locations => 'System-CPython-2.7:/usr/bin/python2.7'
    }

Puppet complains that:

    "-toolLocations" is not a valid option

jhoblitt:
What version of the module and puppet are you using?

jhoblitt:
Could you post the complete error message?  I suspect the error your seeing is not coming from puppet itself.

patcadelina:
Versions:

    // from metadata.json
    "name": "rtyler-jenkins",
    "version": "1.5.0",

    $puppet --version
    2.7.11

Error message:

    "-toolLocations" is not a valid option
     -autoDiscoveryAddress VAL      : Use this address for udp-based auto-discovery
                                      (default 255.255.255.255)
     -candidateTag VAL              : Show swarm candidate with tag only
     -description VAL               : Description to be put on the slave
     -disableClientsUniqueId        : Disables Clients unique ID.
     -disableSslVerification        : Disables SSL verification in the HttpClient.
     -executors N                   : Number of executors
     -fsroot FILE                   : Directory where Jenkins places files
     -help (--help)                 : Show the help screen
     -labels VAL                    : Whitespace-separated list of labels to be
                                      assigned for this slave. Multiple options are
                                      allowed.
     -master VAL                    : The complete target Jenkins URL like
                                      'http://server:8080/jenkins/'. If this option
                                      is specified, auto-discovery will be skipped
     -mode MODE                     : The mode controlling how Jenkins allocates
                                      jobs to slaves. Can be either 'normal'
                                      (utilize this slave as much as possible) or
                                      'exclusive' (leave this machine for tied jobs
                                      only). Default is normal.
     -name VAL                      : Name of the slave
     -noRetryAfterConnected         : Do not retry if a successful connection gets
                                      closed.
     -password VAL                  : The Jenkins user password
     -passwordEnvVariable VAL       : Environment variable that the password is
                                      stored in
     -retry N                       : Number of retries before giving up. Unlimited
                                      if not specified.
     -showHostName (--showHostName) : Show hostnames instead of IP address
     -t (--toolLocation)            : A tool location to be defined on this slave.
                                      It is specified as 'toolName=location'
     -tunnel VAL                    : Connect to the specified host and port,
[...]

PR #367, opened 2015-09-04
[puppet-jenkins#366] Replace -toolLocations with --toolLocation

patcadelina:
Should fix issue #366

### DIFF
--- a/manifests/slave.pp
+++ b/manifests/slave.pp
@@ -95,6 +95,8 @@ class jenkins::slave (
   $enable                   = true
 ) inherits jenkins::params {
 
+  validate_string($tool_locations)
+
   $client_jar = "swarm-client-${version}-jar-with-dependencies.jar"
   $client_url = "http://maven.jenkins-ci.org/content/repositories/releases/org/jenkins-ci/plugins/swarm-client/${version}/"
 
--- a/spec/classes/jenkins_slave_spec.rb
+++ b/spec/classes/jenkins_slave_spec.rb
@@ -37,6 +37,13 @@ describe 'jenkins::slave' do
       let(:params) { {:enable => false, :ensure => 'stopped' } }
       it { should contain_service('jenkins-slave').with(:enable => false, :ensure => 'stopped') }
     end
+
+    describe 'with tool_locations' do
+      let(:params) { { :tool_locations => 'Python-2.7:/usr/bin/python2.7 Java-1.8:/usr/bin/java' } }
+      it { should contain_file(slave_runtime_file).
+        with_content(/--toolLocation Python-2.7=\/usr\/bin\/python2.7/).
+        with_content(/--toolLocation Java-1.8=\/usr\/bin\/java/) }
+    end
   end
 
   shared_examples 'using slave_name' do
--- a/templates/jenkins-slave-defaults.erb
+++ b/templates/jenkins-slave-defaults.erb
@@ -50,8 +50,6 @@ CLIENT_NAME="<%= @slave_name -%>"
 
 FSROOT="<%= @slave_home -%>"
 
-TOOL_LOCATIONS="<%= @tool_locations -%>"
-
 DESCRIPTION="<%= @description -%>"
 
 # credentials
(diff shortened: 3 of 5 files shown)


manifests/slave.pp
```diff
@@ -49,7 +49,7 @@
 #   Not required.  Single string of whitespace-separated list of labels to be assigned for this slave.
 #
 # [*tool_locations*]
-#   Not required.  Single string of whitespace-separated list of tool locations to be defined on this slave. A tool location is specified as 'toolName:location'.
+#   Not required.  Array list of tool locations to be defined on this slave. A tool location is specified as 'toolName:location'.
```
rtyler: This does change an API which concerns me, it would be better to take an Array or the single String for the parameter and then prepare the appropriate data for the template

patcadelina:
Updated.

rtyler:
Without some [rspec-puppet](http://rspec-puppet.com) tests this will have to wait until I have a chance to write some, which will likely be in the next couple weekends

patcadelina:
I went ahead to write one. Also, from the docs,

    -t (--toolLocation)            : A tool location to be defined on this slave.
                                  It is specified as 'toolName=location'

Not sure if we have to go with `--toolLocation toolName=location`. Right now this PR resolves to `--toolLocation toolName:location` as we are directly referencing `slave::tool_locations`.

> [ tool_locations ]
> Not required.  Single string of whitespace-separated list of tool locations to be defined on this slave. A tool location is specified as 'toolName:location'.

How do you intend to test which works?

jhoblitt:
I have not used the `--toollocation` flag with swarm, so I can't comment on what the correct calling semantics are.

The string splitting logic introduced into `templates/jenkins-slave-defaults.erb` is not covered by the spec tests.  It might also be a good idea to validate the `toollocation` param as either `undef` or `string` with `validate_string()` to catch an array being passed by accident.  Another possibility would be to support both a white space delimited string or an array of options.

jhoblitt:
A hash might be better than an array if it's worth the effort to support multiple types.  Would there ever be a use case to build up a list of a tool paths from hiera?

patcadelina:
I tested for the right syntax, updated the diff for this as well.

    $ java -jar swarm-client-2.0-jar-with-dependencies.jar --toolLocation foo:bar
    An argument for setting a Map must contain a "="
    $ java -jar swarm-client-2.0-jar-with-dependencies.jar --toolLocation foo=bar
    Discovering Jenkins master

I'm not sure about hiera, haven't used it myself.

spec/classes/jenkins_slave_spec.rb
```diff
@@ -37,6 +37,16 @@
...
     end
+
+    describe 'with single tool location' do
+      let(:params) { { :tool_locations => 'Python-2.7:/usr/bin/python2.7' } }
+      it { should contain_file(slave_runtime_file).with_content(/--toolLocation Python-2.7=\/usr\/bin\/python2.7/) }
+    end
+
+    describe 'with multiple tool locations' do
+      let(:params) { { :tool_locations => 'Python-2.7:/usr/bin/python2.7 Java-1.8:/usr/bin/java' } }
+      it { should contain_file(slave_runtime_file).with_content(/--toolLocation Python-2.7=\/usr\/bin\/python2.7|--toolLocation Java-1.8=\/usr\/bin\/java/) }
```
jhoblitt: This doesn't test that both option string are present.  How about something along these lines?

```
it do
  should contain_file(slave_runtime_file).
    with_content(/--toolLocation Python-2.7=\/usr\/bin\/python2.7/).
    with_content(/--toolLocation Java-1.8=\/usr\/bin\/java/)
end
```

patcadelina:
Thanks for the feedback.

templates/jenkins-slave-defaults.erb
```diff
@@ -80,9 +78,12 @@ if [ -n "$FSROOT" ]; then
   FSROOT_ARG="-fsroot '$FSROOT'"
 fi
 
-if [ -n "$TOOL_LOCATIONS" ]; then
-  TOOL_LOCATIONS_ARG="-toolLocations '$TOOL_LOCATIONS'"
-fi
+<% if @tool_locations %>
+  TOOL_LOCATIONS_ARG=""
+  <% @tool_locations.split.each do |tool_location| %>
+    TOOL_LOCATIONS_ARG='$TOOL_LOCATIONS_ARG --toolLocation <%= tool_location.split(':').first -%>=<%= tool_location.split(':').last -%>'
```
jhoblitt: What if the string is already in the `Python-2.7=/usr/bin/python2.7` format?  If there are any existing users of the `tool_location` param, wouldn't they have to be using that form?

patcadelina:
@rtyler was concerned about breaking the API. I could go either way really.

jhoblitt:
It looks like version 2.0 intentionally made a breaking change to the `-toollocation[s]` semantics.

https://github.com/jenkinsci/swarm-plugin/pull/28

I'd vote for supporting the current plugin version since, as-is, user's will experience breakage when upgrading swarm

rtyler:
Great catch @jhoblitt! I can buy that, we'll just have to document it thoroughly, sorry for putting this through such scrutiny @patcadelina

jhoblitt:
A validation regex should be able to catch the old format with a useful warn/fail message.

patcadelina:
No worries.

jhoblitt:
@patcadelina - I'd like to echo @rtyler and thank you for your persistence in enduring such a long review process.

PR #374, opened 2015-09-11
add single quotes for credentials

iakovgan:
also strip inital credentials for single quotes, for backward compatibility

### DIFF
--- a/templates/jenkins-slave-defaults.erb
+++ b/templates/jenkins-slave-defaults.erb
@@ -54,9 +54,9 @@ TOOL_LOCATIONS="<%= @tool_locations -%>"
 
 DESCRIPTION="<%= @description -%>"
 
-# credentials
-JENKINS_USERNAME="<%= @ui_user -%>"
-JENKINS_PASSWORD="<%= @ui_pass -%>"
+# credentials should be single quoted
+JENKINS_USERNAME="'<%= @ui_user.gsub!(/^\'|\'?$/, '') -%>'"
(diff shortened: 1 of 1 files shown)


templates/jenkins-slave-defaults.erb
```diff
@@ -54,9 +54,9 @@ TOOL_LOCATIONS="<%= @tool_locations -%>"
 
 DESCRIPTION="<%= @description -%>"
 
-# credentials
-JENKINS_USERNAME="<%= @ui_user -%>"
-JENKINS_PASSWORD="<%= @ui_pass -%>"
+# credentials should be single quoted
+JENKINS_USERNAME="'<%= @ui_user.gsub!(/^\'|\'?$/, '') -%>'"
```
jhoblitt: This will raise an exception when `@ui_user` is `nil` (`undef`). 

Why isn't `JENKINS_USERNAME='<%= @ui_user -%>'` sufficient to stop the shell from interpolating the string?

jhoblitt:
A possible alternative solution would be to [`shellquote()`](https://docs.puppetlabs.com/references/latest/function.html#shellquote).

templates/jenkins-slave-defaults.erb
iakovgan: Yes you right exception on undef is really bad.
 
This regexp supposed to remove quotes if user input is already using quoted value as a workaround. So we could prevent double quoting.

jhoblitt:
@iakovgan This PR was fixed up and merged as #395.  Thank you for contributing!

PR #382, opened 2015-10-02
(RFC) native types and providers

jhoblitt:
This is a family of native types and providers with the aim of replacing direct usage of the `cli` jar via `exec` resources (E.g. `jenkins::cli:exec`).  The principle motivation is to resolve the "chicken vs egg" problem that occurs between the `cli` jar and jenkins when enabling security.  Similarly, it provides a mechanism to set the API token on a user account that can be used for the connection of swarm clients, which otherwise are unable to connect when using a security realm that does not use password (E.g. github-oauth).  It should be relatively easy to extend these types to provide additional functionality, such as supporting additional credentials types.

No attempt has been made to integrate the new types into the existing classes and defined types.  It should be possible to try them out in an essentially stand alone fashion against a running jenkins instance.  Preliminary and extremely terse notes are in [NATIVE_TYPES_AND_PROVIDERS.md](https://github.com/jhoblitt/puppet-jenkins/blob/feature/types_and_providers/NATIVE_TYPES_AND_PROVIDERS.md).

Comments, feedback, and/or bug reports would be greatly welcomed.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -17,6 +17,8 @@ group :development do
   gem 'debugger', :platform => :mri_19
   gem 'debugger-pry', :platform => :mri_19
   gem 'byebug', :platform => [:mri_20, :mri_21]
+  gem 'pry'
+  gem 'pry-byebug'
 end
 
 group :system_tests do
--- a/Gemfile
+++ b/Gemfile
@@ -9,6 +9,7 @@ gem 'puppet-syntax', '>= 1.1.0'
 gem 'json'
 gem 'puppet', ENV['PUPPET_VERSION'] || '~> 3.5'
 gem 'metadata-json-lint'
+gem 'retries', '~> 0.0.5'
 
 group :development do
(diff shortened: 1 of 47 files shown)


rtyler:
I think this is worth including. i'm not sure, past defaulting these behaviors, how to encourage their usage though

jhoblitt:
@rtyler I agree that advertising this will be difficult.  Perhaps add a section to the README about experimental types?

There is a [todo](https://github.com/jenkinsci/puppet-jenkins/blob/master/NATIVE_TYPES_AND_PROVIDERS.md#todo) in my usage notes for known issues.  In particular, I think there needs to be some discussion about the `enabled` param to `jenkins::job` as it may break idempotency by modifying the job's `config.xml`.  The easiest thing to do would be to deprecate the param and make it a "no-op".  Another option would be to ignore managing the job's config string when the parameter is set.

PR #386, opened 2015-10-08
set_security() does not save jenkins state

### LINKED ISSUE #223: Jenkins security setting is lost after service restart
I have configured my jenkins instance in ubuntu 12.04 using

```puppet
  class { '::jenkins::security':
    security_model => 'full_control',
  }
```

which sets the settings to use user/password correctly, but  if I restart the service these settings are lost, giving the world full access without user/password...

Is this by design?

jhoblitt:
Resolves #223.


4m3ndy:
this bug still exists in jenkins 2.19.4 ? any recommendations ?

PR #389, opened 2015-10-10
remove seperate resources for handling plugin extension

### LINKED ISSUE #380: Plugins not pinned and updated when using jenkins class
I have hieradata containing plugin names and versions which I pass to jenkins::plugin type in the same class as I call the jenkins class to install a version of Jenkins. However, I have to run the script twice, as the second run is needed to pin plugins from the bundled version of jenkins and to install the ones I have specified in my hieradata.

Perhaps I am not using the module as intended? Please advise

jhoblitt:
There were two orthogonal resource chains for handling plugins with with
the .hpi & .jpi extension.  This commit unifies the two code paths.  It
should fix the resource ordering problem that causes it to take two runs
to install a plugin and pin it.

resolves #380

### DIFF
--- a/manifests/plugin.pp
+++ b/manifests/plugin.pp
@@ -32,7 +32,6 @@ define jenkins::plugin(
 ) {
   include ::jenkins::params
 
-  $plugin            = "${name}.hpi"
   $plugin_parent_dir = inline_template('<%= @plugin_dir.split(\'/\')[0..-2].join(\'/\') %>')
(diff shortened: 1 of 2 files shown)


jhoblitt:
Alternative to #388 that is a bit more aggressive.

jhoblitt:
I suspect the pin file should be removed when the plugin is `:absent`...

PR #390, opened 2015-10-10
Adds Examples for various platforms for Jenkins

jhoblitt:
Examples are moved from tests/ -> examples as per https://docs.puppetlabs.com/puppet/latest/reference/modules_fundamentals.html#module-layout


jhoblitt:
This fixes the linter issues with #346

PR #396, opened 2015-10-11
add user/group mgt. + localstatedir params to jenkins class

### LINKED ISSUE #219: Plugins are installed each time and restarting service
Every time puppet runs, the plugin installation is run even when it already exists, causing a service restart each time...

### LINKED ISSUE #249: Path to $libdir hardcoded in few manifests
We needed to deploy Jenkins into alternate location on the file system. To achieve this we had to set the custom $config_hash = { 'JENKINS_HOME' => { 'value' => "$jenkins::params::jenkins_home" },
and different value $jenkins::params::libdir. 

However, following modules have a hard coded path to /var/lib/jenkins/. 
/plugin.pp
/package.pp
/proxy.pp
/job/present.pp
/job/absent.pp

Suggesting a change to allow for use of  $jenkins::params::libdir across all modules.

Also the dedicated $jenkins::params::jenkins_home which would not require $config_hash would be nice.

Robert

### LINKED ISSUE #250: Somehow a change to a plugin is restarting jenkins quite often
I have the version of the plugin fixed in the puppet manifest:

```
    jenkins::plugin{ 'ghprb':
        version => '1.16-8'
    }
```

But I will see this happen on occasion. I don't know why, because I didn't delete that plugin, or anything else. I don't know why it's getting the plugin again, because it shouldn't be. But this is causing our jenkins to restart in some cases every 20 minutes....

```
Feb 13 13:07:49 jenkins-proto puppet-agent[12424]: (/Stage[main]/Repose_jenkins::Master/Jenkins::Plugin[ghprb]/Exec[download-ghprb]/re
turns) executed successfully
Feb 13 13:07:49 jenkins-proto puppet-agent[12424]: (/Stage[main]/Repose_jenkins::Master/Jenkins::Plugin[ghprb]/File[/var/lib/jenkins/p
lugins/ghprb.hpi]/owner) owner changed 'root' to 'jenkins'
Feb 13 13:07:51 jenkins-proto jenkins: jenkins: client (pid 9935) exited with 143 status
Feb 13 13:07:52 jenkins-proto puppet-agent[12424]: (/Stage[main]/Jenkins::Service/Service[jenkins]) Triggered 'refresh' from 1 events
Feb 13 13:07:52 jenkins-proto puppet-agent[12424]: (/Stage[main]/Base/Service[chrony]/ensure) ensure changed 'stopped' to 'running'
Feb 13 13:07:52 jenkins-proto puppet-agent[12424]: Finished catalog run in 5.94 seconds
```

### LINKED ISSUE #362: Error on updating existing job
When I am trying to update an existing jenkins job with this module, I get an error during the agent run.

Code:
```ruby
jenkins::job { 'bounce-apache':
    config => template('ha_jenkins_sandbox/bounce-apache/config.xml.erb'),
}
```

Error:
```
Info: Applying configuration version '766ab92'
Error: cat "/tmp/bounce-apache-config.xml" | java -jar /usr/lib/jenkins/jenkins-cli.jar -s http://localhost:8081 create-job "bounce-apache" returned 255 instead of one of [0]
Error: /Stage[main]/Ha_jenkins_sandbox/Jenkins::Job[bounce-apache]/Jenkins::Job::Present[bounce-apache]/Exec[jenkins create-job bounce-apache]/returns: change from notrun to 0 failed: cat "/tmp/bounce-apache-config.xml" | java -jar /usr/lib/jenkins/jenkins-cli.jar -s http://localhost:8081 create-job "bounce-apache" returned 255 instead of one of [0]
Notice: /Stage[main]/Ha_jenkins_sandbox/Jenkins::Job[bounce-apache]/Jenkins::Job::Present[bounce-apache]/Exec[jenkins enable-job bounce-apache]: Dependency Exec[jenkins create-job bounce-apache] has failures: true
Warning: /Stage[main]/Ha_jenkins_sandbox/Jenkins::Job[bounce-apache]/Jenkins::Job::Present[bounce-apache]/Exec[jenkins enable-job bounce-apache]: Skipping because of failed dependencies
```

```
cat "/tmp/bounce-apache-config.xml" | java -jar /usr/lib/jenkins/jenkins-cli.jar -s http://localhost:8081 create-job "bounce-apache"
```
When I run the above command it complains that the job already exists and exits with the 255 return code.   If I delete the job and rerun this issue goes away but comes back for the next agent checkin.  Looking at the jenkins::job file there seems to be an update-job but it is never called for me as the run fails after create is called.

Am I missing a step to avoid calling create-job on every job?  A majority of the time all of my jobs will already exist so this seems pretty broken.

### LINKED ISSUE #365: jenkins user and jenkins_home directory not configurable
Hi,

It seems that the jenkins main manifest to install jenkins master, creates a OS level user named "jenkins" with a home directory of "/var/lib/jenkins". The OS user and home directory are not configurable. I believe this has been done because the yum package to install Jenkins, always installs the master to the /var/lib/jenkins directory.

In my scenario, I have a jenkins user that I have created myself prior to running this manifest which has a home directory of /home/jenkins. Jenkins master is still installed in /var/lib/jenkins, however, this means that jenkins plugins are reinstalled everytime the puppet script is run, as lib/puppet/jenkins.rb always looks in "~jenkins" to find the plugins directory, which in my case does not exist. 

My change request would be to allow the OS user and its home directory, as well as the Jenkins home directory to be configurable, so that these can be set by the module user.

Thanks

jhoblitt:
* add `manager_user`, `user`, `manage_group`, `group` params to
  `jenkins` class.

* improve consistency of `file` resouce group ownership throughout
  the module

* add 'localstatedir' param to `jenkins` class and consistently use it
  to set the 'data' base dir throughout the module

* Deprecate/noop `username`, `group`, `create_user` params to
  `jenkins::plugin` define; replaced by `user`,`group`,etc. params to
  `jenkins` class

* Deprecate/noop `plugin_dir` param to `jenkins::plugin`; replaced by
  `localstatedir` param to `jenkins` class

closes #356 -- alternative approach
resolves #362
resolves #365
resolves #219 
resolves #249
resolves #250

### DIFF
--- a/manifests/cli_helper.pp
+++ b/manifests/cli_helper.pp
@@ -20,8 +20,8 @@ class jenkins::cli_helper (
   $helper_groovy = "${libdir}/puppet_helper.groovy"
   file {$helper_groovy:
     source  => 'puppet:///modules/jenkins/puppet_helper.groovy',
-    owner   => 'jenkins',
-    group   => 'jenkins',
+    owner   => $::jenkins::user,
+    group   => $::jenkins::group,
     mode    => '0444',
     require => Class['jenkins::cli'],
   }
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -1,14 +1,4 @@
-# Parameters:
-# config_hash = {} (Default)
-# Hash with config options to set in sysconfig/jenkins defaults/jenkins
-#
-# Example use
-#
-# class{ 'jenkins::config':
-#   config_hash => {
-#     'HTTP_PORT' => { 'value' => '9090' }, 'AJP_PORT' => { 'value' => '9009' }
-#   }
-# }
+# This class should be considered private
 #
 class jenkins::config {
 
@@ -17,4 +7,35 @@ class jenkins::config {
   }
 
   create_resources( 'jenkins::sysconfig', $::jenkins::config_hash )
+
+  $dir_params = {
+    ensure => directory,
+    owner  => $::jenkins::user,
+    group  => $::jenkins::group,
+    mode   => '0755',
+  }
+
+  # ensure_resource is used to try to maintain backwards compatiblity with
+  # manifests that were able to external declare resources due to the
+  # old conditional behavior of jenkins::plugin
+  if $::jenkins::manage_user {
+    ensure_resource('user', $::jenkins::user, {
+      ensure     => present,
+      gid        => $::jenkins::group,
+      home       => $::jenkins::localstatedir,
+      managehome => false,
+      system     => true,
+    })
+  }
+
+  if $::jenkins::manage_group {
+    ensure_resource('group', $::jenkins::group, {
+      ensure => present,
+      system => true,
+    })
+  }
+
+  ensure_resource('file', $::jenkins::localstatedir, $dir_params)
+  ensure_resource('file', $::jenkins::plugin_dir, $dir_params)
+  ensure_resource('file', $::jenkins::job_dir, $dir_params)
 }
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -44,11 +44,25 @@
 # config_hash = undef (Default)
 #   Hash with config options to set in sysconfig/jenkins defaults/jenkins
 #
+# localstatedir = '/var/lib/jenkins' (default)
+#   base path, in the autoconf sense, for jenkins local data including jobs and
+#   plugins
+#
 # executors = undef (Default)
 #   Integer number of executors on the Jenkin's master.
 #
 # slaveagentport = undef (Default)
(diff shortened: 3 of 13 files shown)


jhoblitt:
@rtyler This should resolve #365 and is a more intrusive alternative to #356 (which I request to be split out of #304).  I'm slightly afraid of breakage so perhaps it should wait until after the 1.6.0 release?

jhoblitt:
I would have prefer to split this into to separate PRs but the user management change is tightly coupled to directory management.

jhoblitt:
Looks like this should resolve #362 as well.

rtyler:
Nah, let's move fast and break things! :+1:

PR #398, opened 2015-10-13
client_url is hardcoded in slave.pp

zonArt:
This PR intends to give the possibility to change it

### DIFF
--- a/manifests/slave.pp
+++ b/manifests/slave.pp
@@ -94,7 +94,6 @@ class jenkins::slave (
   $ensure                   = 'running',
   $enable                   = true,
   $source                   = undef,
-  $maven_repo_url           = 'http://maven.jenkins-ci.org/content/repositories/releases',
 ) inherits jenkins::params {
 
   validate_string($tool_locations)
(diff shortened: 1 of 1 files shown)


manifests/slave.pp
```diff
@@ -93,12 +93,13 @@
   $install_java             = $jenkins::params::install_java,
   $ensure                   = 'running',
   $enable                   = true
```
jhoblitt: missing comma

manifests/slave.pp
```diff
@@ -93,12 +93,13 @@
...
+  $maven_repo_url           = 'http://maven.jenkins-ci.org/content/repositories/releases',
```
jhoblitt: It's probably not important that the download path is actually a maven repo.  Would it be better to have a param to specify the complete url to the jar?

zonArt: done

jhoblitt:
@zonArt Could you squash these commits down and give us a headsup when you feel this PR is ready?

rtyler:
@jhoblitt IMHO squashing commits isn't necessary, it's often nice but meh

PR #400, opened 2015-10-14
fix job_dir variable reference to jenkins class

shieldwed:
This seems to be just a minor typo, but it causes create job exec statements to fail the second time, since puppet would retry creating jobs. It cannot find the correct files because of the incomplete path.


rtyler:
Gak! Good catch, I'll release a 1.6.1 with this fix soon

jhoblitt:
We obviously have test coverage issues. :frowning:

Issue #403, opened 2015-10-17
Ensure $::jenkins::localstatedir to Directory Breaks Filesystem's With Symlinked Mounts


malnick:
On our deployment we mount an EBS volume in and symlink to /var/lib/jenkins. Ensuring this path to a directory in the module breaks that symlink. This is a sane default but could be improved by exposing $::jenkins::localstatedir file-type as a tunable parameter, giving the end-user more flexibility in the type of file that this should be. I'll submit a PR for this feature.

esalberg:
@jhoblitt It looks like this issue can be closed?

PR #404, opened 2015-10-17
Enable setting $localstatedir file type [WIP]

malnick:
Per https://github.com/jenkinsci/puppet-jenkins/issues/403

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -34,8 +34,22 @@ class jenkins::config {
       system => true,
     })
   }
+  # Allow users to set jenkins home to a link in order to allow for NFS or 
+  # other remote file mounts to symlink to the directory. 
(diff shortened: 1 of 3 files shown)


jhoblitt:
@malnick My general preference is to allow functionality to be turned off so it can be managed externally rather than trying to build up the logic for all use cases.  How about a param to simply disable directory management, making it the end user's responsibility?

malnick:
@jhoblitt That sounds fine to me! Let me fix this PR to reflect that.

chizou:
How does this PR meet what @jhoblitt mentioned? I don't see any method to disable management of the home directory. I'm with him that permutations, like having a symlink, should be managed by an external source and we this module should simply enable/disable management of the home directory.

PR #405, opened 2015-10-17
Enable toggle to manage $localstatedir (Jenkins Home)

malnick:

PR #407, opened 2015-10-22
Allow the user to manage the localstatedir themselves.

jniesen:
* add `manage_localstatedir` param to `jenkins class
* align params on equal sign.
* validate that manage_localstatedir is a bool.
* add documentation for manage_localstatedir.

When localstatedir is a symlink to a mount, the user may want to manage that outside of this module.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -35,9 +35,6 @@ class jenkins::config {
     })
   }
 
-  if $::jenkins::manage_localstatedir {
-  }
-
   if $::jenkins::manage_datadirs {
     ensure_resource('file', $::jenkins::localstatedir, $dir_params)
     ensure_resource('file', $::jenkins::plugin_dir, $dir_params)
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -35,7 +35,10 @@ class jenkins::config {
     })
   }
 
-  ensure_resource('file', $::jenkins::localstatedir, $dir_params)
+  if $::jenkins::manage_localstatedir {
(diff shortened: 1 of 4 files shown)


manifests/config.pp
```diff
@@ -35,7 +35,10 @@
     })
   }
 
-  ensure_resource('file', $::jenkins::localstatedir, $dir_params)
+  if $::jenkins::manage_localstatedir {
+    ensure_resource('file', $::jenkins::localstatedir, $dir_params)
+  }
+
```
jhoblitt: @jniesen A similar PR, #405, disabled all of the directory resources.  Do you have any opinion as to which way to go?

jniesen: @jhoblitt In my case, my localstatedir already contains those directors and they're managed by the user and group that I'm passing in as the jenkins user and group, so it wouldn't change much. I was more just worried about having the option of defining localstatedir as a symlink.

It might be a good idea to toggle all the directory resources to support other possibilities.

jhoblitt: That seems reasonable to me.  Are you willing to update the PR?

jniesen: I will update the PR so plugin dir and job dir are in the same conditional block as localstatedir. It'll be in a couple of hours.

jniesen:
@jhoblitt Is this what you had in mind?

jhoblitt:
@jniesen I can't remember now why I didn't immediately merge this... so I'm going to go ahead and merge it.  Thank you for contributing!

PR #411, opened 2015-10-24
Check if the retries gem is not already declared

boyand:
The main motivation for this is handling the retries gem installation when it is already defined in the catalog. 
This is needed if one has to use a different provider. For example for Puppet 4 to install the retries gem successfully one needs to use the 'puppet_gem' provider. 
Alternative solution would be to handle this in the module directly.

### DIFF
--- a/manifests/cli/config.pp
+++ b/manifests/cli/config.pp
@@ -24,8 +24,10 @@ class jenkins::cli::config(
   validate_string($ssh_private_key_content)
 
   # required by PuppetX::Jenkins::Provider::Clihelper base
-  package { 'retries':
-    provider => 'gem',
+  if ! defined(Package['retries']) {
+    package { 'retries':
+      provider => 'gem',
+    }
   }
 
   if $ssh_private_key and $ssh_private_key_content {


manifests/cli/config.pp
```diff
@@ -24,8 +24,10 @@
   validate_string($ssh_private_key_content)
 
   # required by PuppetX::Jenkins::Provider::Clihelper base
-  package { 'retries':
-    provider => 'gem',
+  if ! defined(Package['retries']) {
```
rtyler: We've been making these changes elsewhere, it's probably better to just use [ensure_packages](https://github.com/puppetlabs/puppetlabs-stdlib#ensure_packages) for this.

If you've not got the time, I can update the changeset tomorrow

boyand: I usually try to resolve "generic" package issues using ensure_packages as well. However it doesn't work in this specific case. The challenge here is that one needs to use a different provider for this package. The following results in a duplicate declaration because it cannot be merged properly:
```
#config.pp
ensure_packages(['retries'],{'provider' => 'gem'})
```
```
#other.pp
ensure_packages(['retries'],{'provider' => 'puppet_gem'})
```

jhoblitt: @boyand There has been some discussion of puppet "slotting" resources by provider specifically because of the 3 different providers for needed for "unvendored", puppetserver, & puppet-agent.  I'm not sure what the correct thing to do here is -- perhaps remove the resource completely and make it the end-user's responsibility?

boyand: @jhoblitt it is not an ideal situation given the different distributions. If we use 'defined' it should at least be backwards compatible. Then again given that this affects the experimental types only, we may just remove it altogether and make it easier for everyone going forward.

PR #412, opened 2015-10-26
Split responsibility between configuration file creation and associated job creation

zonArt:


jhoblitt:
What is the use case for this change?

zonArt:
It permits to have the source of the configuration file coming from the disk directly (instead of a puppet resource as it is the case now) and it just splits the responsibilities between job creation and configuration file creation. In our usecase we don't need to create a tmp file as we manage this creation by contcat fragments before using jenkins-cli

jhoblitt:
@zonArt Why are you trying to use `jenkins::job::present` directly instead of the [`source`](https://github.com/jenkinsci/puppet-jenkins/blob/master/manifests/job.pp#L10-L12)  param to `jenkins::job`?

zonArt:
@jhoblitt Yes you're right, sorry I should have digged a little deeper for the possibilities offered. `source` param in `jenkins::job` works perfect. However, why is `config` required if you're using `source` or `template` ?
PR can be declined

PR #414, opened 2015-10-26
Changed diff to xmldiff

zonArt:
making it a little simplier to compare and removing the substitution as it is ok for xmldiff to deal with them correctly, adds a dependency on https://www.logilab.org/project/xmldiff

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -25,5 +25,6 @@
     { "name": "darin/zypprepo", "version_requirement": ">= 1.0.1 < 2.0.0" },
     { "name": "nanliu/staging", "version_requirement": ">= 1.0.0 < 2.0.0" },
     { "name": "camptocamp/archive", "version_requirement": ">= 0.8.0" }
+    { "name": "logilab/xmldiff", "version_requirement": ">=0.6.10" },
   ]
 }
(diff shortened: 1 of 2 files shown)


metadata.json
```diff
@@ -25,5 +25,6 @@
     { "name": "darin/zypprepo", "version_requirement": ">= 1.0.1 < 2.0.0" },
     { "name": "nanliu/staging", "version_requirement": ">= 1.0.0 < 2.0.0" },
     { "name": "camptocamp/archive", "version_requirement": ">= 0.8.0" }
+    { "name": "logilab/xmldiff", "version_requirement": ">=0.6.10" },
```
jhoblitt: There's no such module published on the forge.

zonArt: You're right, I didn't know if the dependencies were only for puppet modules or for every third library. To make this modification a python module is needed, it is called xmldiff (https://pypi.python.org/pypi/xmldiff). What is your suggestion to reflect this needs, modification of the readme ?

jhoblitt:
@zonArt I'm not really keen on installing a package via pip since there may not be a working python env.  Does #418 cover your needs?

zonArt:
@jhoblitt Ok I understand you're point and I agree that it probably isn't the best option. #418 looks good but as the new types are still in beta, I'm not really confident to use them now for our production instance, though I'll take a look at them. At least could I modify this PR to grant the ability to chose the difftool used ?

zonArt:
#422 to replace this one

PR #415, opened 2015-10-26
Added a ^

zonArt:
in order to ensure that we grep on a tag that begins the line (disable tag may also be in the file for other reason)

### DIFF
--- a/manifests/job/present.pp
+++ b/manifests/job/present.pp
@@ -87,7 +87,7 @@ define jenkins::job::present(
   if ($enabled == 1) {
     exec { "jenkins enable-job ${jobname}":
       command => "${jenkins_cli} enable-job \"${jobname}\"",
-      onlyif  => "cat \"${config_path}\" | grep '$<disabled>true'",
+      onlyif  => "cat \"${config_path}\" | grep '^<disabled>true'",
       require => [
         Exec["jenkins create-job ${jobname}"],
         Exec["jenkins update-job ${jobname}"],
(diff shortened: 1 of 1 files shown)


manifests/job/present.pp
```diff
@@ -87,7 +87,7 @@
   if ($enabled == 1) {
     exec { "jenkins enable-job ${jobname}":
       command => "${jenkins_cli} enable-job \"${jobname}\"",
-      onlyif  => "cat \"${config_path}\" | grep '<disabled>true'",
+      onlyif  => "cat \"${config_path}\" | grep '$<disabled>true'",
```
jhoblitt: Did you test this change as working?

jhoblitt: (you may want to check the man page for grep)

zonArt: sorry, I totally messed it up with ^, corrected (and yes it is working)

jhoblitt:
@zonArt Are you using the `enabled` param?  Since it breaks idempotence, I'd like to deprecate it and make it a no-op.

zonArt:
@jhoblitt Yes I do and this PR's purpose is to re-establish the idempotence since as I said this tag is at the beginning of the line so with this symbol, everything works again as it should

zonArt:
@jhoblitt ok I think I know what you mean, this flag (enabled) breaks idempotence unless it is coherent with the XML template you give and at that point you dont need it anymore. Did I got you well ?

jhoblitt:
@zonArt With at least Jenins 1.61x, this won't work because the tag is indented.  I think this "feature" should be removed as it's fundamentally broken.

zonArt:
@jhoblitt: Looks like this "feature" (or maybe "bug") still exists even if you mentioned it would be removed. Could I propose a PR to remove this part ?

PR #417, opened 2015-10-30
Unify the ssh_keyfile

chizou:
Currently, we need to define the ssh_keyfile in two locations: cli.pp and cli_helper.pp. My current hiera yaml file needs both of the following entries to work correctly:

jenkins::cli_helper::ssh_keyfile: '/var/lib/jenkins/.ssh/id_rsa'
jenkins::cli_ssh_keyfile: '/var/lib/jenkins/.ssh/id_rsa'

This change will make the first entry unnecessary.

### DIFF
--- a/manifests/cli_helper.pp
+++ b/manifests/cli_helper.pp
@@ -2,9 +2,7 @@
 #
 # A helper script for creating resources via the Jenkins cli
 #
-class jenkins::cli_helper (
-  $ssh_keyfile = undef,
-){
+class jenkins::cli_helper {
   include ::jenkins
   include ::jenkins::cli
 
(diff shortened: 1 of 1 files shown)


manifests/cli_helper.pp
```diff
@@ -2,9 +2,7 @@
 #
 # A helper script for creating resources via the Jenkins cli
 #
-class jenkins::cli_helper (
-  $ssh_keyfile = undef,
```
rtyler: My preference would be that this argument stays so it's not used, then adding a deprecation warning if it's being used will ensure that we don't break current implementations

rtyler:
I'm going to update this code with some more tests after merging it.

PR #419, opened 2015-11-07
Fix Plugin Download Timeout

mooreandrew:
When downloading large plugins, it can timeout (e.g. ruby-runtime). The
archive::download supports a timeout value which defaults to 120
seconds, but this plugin class doesn’t have any support to override
this value. This allows for the ability to adjust the timeout value for
large plugins.

Error message when download timeout exceeds 120 seconds:

Error:
/Stage[main]/Jenkins::Plugins/Jenkins::Plugin[ruby-runtime]/Archive::Dow
nload[ruby-runtime.hpi]/Exec[download archive ruby-runtime.hpi and
check sum]/returns: change from notrun to 0 failed: Command exceeded
timeout

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,9 @@ If you need to peg a specific version, simply specify that as a string, i.e.:
   }
 ```
 
+Note that plugin will timeout if it takes longer than 120 seconds to download.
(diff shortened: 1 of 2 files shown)


rtyler:
I'm going to go ahead and merge this and add the rspec-puppet necessary myself

PR #420, opened 2015-11-07
Add a test and a little cleanup around jenkins::cli_helper

rtyler:
References #417

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -19,7 +19,7 @@ PuppetLint::RakeTask.new(:lint) do |l|
   l.disable_checks = %w(80chars class_inherits_from_params_class)
   l.ignore_paths = exclude_paths
   l.fail_on_warnings = true
-  l.log_format = "FUK %{path}:%{linenumber}:%{check}:%{KIND}:%{message}"
+  l.log_format = "%{path}:%{linenumber}:%{check}:%{KIND}:%{message}"
 end
 
 PuppetSyntax.exclude_paths = exclude_paths
--- a/manifests/cli_helper.pp
+++ b/manifests/cli_helper.pp
@@ -2,7 +2,16 @@
 #
 # A helper script for creating resources via the Jenkins cli
 #
-class jenkins::cli_helper {
+# Parameters:
+#
+# ssh_keyfile = undef
+#   Defaults to the value of $::jenkins::cli_ssh_keyfile. This parameter is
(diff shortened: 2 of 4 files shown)


Rakefile
```diff
@@ -19,7 +19,7 @@ PuppetLint::RakeTask.new(:lint) do |l|
   l.disable_checks = %w(80chars class_inherits_from_params_class)
   l.ignore_paths = exclude_paths
   l.fail_on_warnings = true
-  l.log_format = "FUK %{path}:%{linenumber}:%{check}:%{KIND}:%{message}"
```
jhoblitt: Sad to see that go...

manifests/cli_helper.pp
```diff
@@ -2,7 +2,16 @@
...
-class jenkins::cli_helper {
+# Parameters:
+#
+# ssh_keyfile = undef
+#   Defaults to the value of $::jenkins::cli_ssh_keyfile. This parameter is
+#   deprecated, please set $::jenkins::cli_ssh_keyfile instead of setting this
+#   directly
+#
+class jenkins::cli_helper (
+  $ssh_keyfile = $::jenkins::cli_ssh_keyfile,
```
jhoblitt: I'm wondering if we need to parameterize this at all? Maybe `$::jenkins::cli_ssh_keyfile` should be the single user entry point and we can work towards making `jenkins::cli_helper` and internal only class?

rtyler: I cannot remove that without breaking API compatibility, but this does remind me  that I was going to put an `info()` message about the parameter being deprecated

manifests/cli_helper.pp
```diff
@@ -25,12 +34,16 @@
...
-  if $::jenkins::cli_ssh_keyfile {
-    $auth_arg = "-i ${::jenkins::cli_ssh_keyfile}"
+  if $ssh_keyfile {
+    $auth_arg = "-i ${ssh_keyfile}"
   } else {
     $auth_arg = undef
   }
 
+  if $ssh_keyfile != $::jenkins::cli_ssh_keyfile {
+    info("Using jenkins::cli_helper($ssh_keyfile) is deprecated and will be removed in the next major version of this module")
```
jhoblitt: `manifests/cli_helper.pp:44:variables_not_enclosed:WARNING:variable not enclosed in {}`

PR #422, opened 2015-11-09
Feature/ability to chose difftool

zonArt:


manifests/job.pp
```diff
@@ -43,6 +44,9 @@
       jobname => $jobname,
     }
   } else {
+    if $config == undef and $source == undef and $template == undef {
+      fail('You should at least specify one of the $config, $source or $template param')
+    }
```
zonArt: Why is this `if` statement (which is obviously useful and avoids misleading configuration) making your rspec tests fail ? Sorry I'm not really used to making rspec tests so I think I will not be able to modify them the right way

manifests/job.pp
```diff
@@ -25,12 +25,13 @@
 #     choose 'absent' to ensure the job is removed
 #
 define jenkins::job(
-  $config,
+  $config   = undef,
```
rtyler: This changes the API of this defined type and makes `$config` optional when it's supposed to be required.

manifests/job.pp
rtyler: In the RSpec (`spec/defines/jenkins_job_spec.rb`) many of the examples specify parameters such as:

 ```ruby
let(:params) {{ :enabled => 0 , :config => '' }}
```

This is basically passing an empty string through as the `$config` parameter, I suppose Puppet is treating an empty string as an `undef`.

rtyler:
I'm cleaning these changes up a bit so they can be merged, will open a new pull request with them shortly.

manifests/job.pp
zonArt: I truly don't understand this approach, what we need here (in this defined type) is to have either `$config` or `$source` or `$template` to be defined (mandatory) and it can be tested (as per my proposal) and make the compilation fail if not respected. What we got here is a constraint on a required parameter which is not used in case we pass either `$source` or `$template`, so why has it to be kept as required ?

manifests/job.pp
zonArt: That's quite strange to have a mandatory parameter not used when the job is disabled

zonArt:
Replaced by #430. I'll open a more global one for my proposal on the `$config` parameter

PR #423, opened 2015-11-09
Gives the possibility to parametrize the  $tmp_config_path

zonArt:
Gives the possibility to parametrize the  $tmp_config_path and avoid tmp config file creation

### DIFF
--- a/manifests/job/present.pp
+++ b/manifests/job/present.pp
@@ -5,7 +5,10 @@
 # Parameters:
 #
 #   config
-#     the content of the jenkins job config file (required)
+#     the content of the jenkins job config file (template based)
+#
+#   tmp_config_path
+#     the jenkins job config file (file on disk)
 #
 #   jobname = $title
 #     the name of the jenkins job
@@ -14,7 +17,8 @@
 #     if the job should be enabled
 #
 define jenkins::job::present(
-  $config,
+  $config = undef,
+  $tmp_config_path = undef,
   $jobname  = $title,
   $enabled  = 1,
   $difftool = '/usr/bin/diff -b -q',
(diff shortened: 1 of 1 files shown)


manifests/job/present.pp
```diff
@@ -14,19 +17,51 @@
...
+    fail('You cannot set both \$temp_config_path AND $config param, only one is required')
+  }
+
   if $jenkins::service_ensure == 'stopped' or $jenkins::service_ensure == false {
     fail('Management of Jenkins jobs requires \$jenkins::service_ensure to be set to \'running\'')
   }
 
   $jenkins_cli        = $jenkins::cli::cmd
-  $tmp_config_path    = "/tmp/${jobname}-config.xml"
+  if $config == undef {
```
zonArt: I'd like to add an `if` statement here to test that we have either `$config` or `$temp_config_path` set but it makes the rpsec tests fail. Put a similar comment in #422

manifests/job/present.pp
```diff
@@ -5,7 +5,10 @@
 # Parameters:
 #
 #   config
-#     the content of the jenkins job config file (required)
+#     the content of the jenkins job config file (template based)
+#
+#   temp_config_path
```
rtyler: If I'm understanding the problem correctly here, the goal is to make the intermediate file used for configuring the job able to be overridden right?

If that's the case, I'm trying to think of a parameter name that would make that more clear so a user wouldn't be confused as to the purpose of the parameter, without knowledge of how the defined type works internally :/

zonArt: yes, that's the idea, we used to call this `$generated` in our module so it could be something like `$generated_config_file`. I used `$temp_config_path` in reference to `$tmp_config_path` you're using hardcoded

zonArt:
Will open something new as this one is in conflict and module evolved, need to rethink that

PR #425, opened 2015-11-09
Make Java Memory and Args Configurable

zutherb:
We had some trouble because the Jenkins Memory is not big enough that why i made it configurable.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 log/
 .project
+.idea
 .blimpy.d
 .vagrant
 pkg/*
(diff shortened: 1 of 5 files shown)


zonArt:
Why don't you use the `config_hash` param ? It looks like perfect for your needs and as this `JAVA_ARGS` key is commented in the template, it will avoid having it activated everywhere (unless the default is that it should have an Xmx at 256m)

zutherb:
hi @zonArt  thx that works. unfortunately i did not see it.

rtyler:
Thanks for jumping in @zonArt, I'm going to close this out now

PR #428, opened 2015-11-12
Suppress notice messages when using archive::download

queeno:
The `verbose` attribute of the `archive::download` resource defaults to `true`.

This causes a notice message, `no checksum for this archive`, to be spammed to the terminal when `archive::download` is called without setting the checksum attribute.

This PR suppresses this notice.


jhoblitt:
:+1:

PR #430, opened 2015-11-15
Minor clean up of difftool patch

rtyler:
See original PR: #422

### DIFF
--- a/manifests/job.pp
+++ b/manifests/job.pp
@@ -24,6 +24,9 @@
 #   ensure = 'present'
 #     choose 'absent' to ensure the job is removed
 #
+#   difftool = '/usr/bin/diff -b-q'
+#     Provide a command to execute to compare Jenkins job files
+#
 define jenkins::job(
   $config,
   $source   = undef,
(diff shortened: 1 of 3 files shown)


manifests/job.pp
```diff
@@ -31,6 +31,7 @@
   $jobname  = $title,
   $enabled  = 1,
   $ensure   = 'present',
+  $difftool = 'diff -b -q',
```
jhoblitt: The param should have some sort of validation, probably `validate_string()`.

manifests/job/present.pp
```diff
@@ -17,6 +17,7 @@
   $config,
   $jobname  = $title,
   $enabled  = 1,
+  $difftool = 'diff -b -q',
```
jhoblitt: Ditto

manifests/job/present.pp
```diff
@@ -78,7 +79,7 @@
   exec { "jenkins update-job ${jobname}":
     command => "${cat_config} | ${update_job}",
     onlyif  => "test -e ${config_path}",
-    unless  => "diff -b -q ${config_path} ${tmp_config_path}",
+    unless  => "${difftool} ${config_path} ${tmp_config_path}",
```
jhoblitt: A doc note suggesting fully qualifying the path to the diffing utility might not be a bad idea.  I can imagine issues being opened over a utility installed in `/usr/local/bin`/etc. not working.

jhoblitt:
This gives me a lot less heartburn than adding a pip dependency. I do have some concern about replicating this semantic in a provider but I guess we could punt and remove it as part of a major API bump when the time comes.

rtyler:
Thanks for the review @jhoblitt, I'll take a crack at cleaning this up before the end of the day

jhoblitt:
:+1:

PR #431, opened 2015-11-17
More flexibility for setup

hdoedens:
removed 'download' from plugin base_url to be more flexible. We have, for example, an artifactory repository as mirror for both jenkins.war files and plugins. The plugin files are in the path: http://servername/artifactory/jenkins-update/plugins. So no download part here.
Mind you that the fix is not backward compatible. I could do this if you want.
Current implementation breaks the Jenkins installation when a symbolic link is used for the plugins or jobs folder. In such case the symbolic link is removed and replaced by an actual folder, effectively removing all jobs from Jenkins \o/.
I added a parameter to follow links in config for setup to enable links for jobs and plugins folder.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -13,6 +13,7 @@ class jenkins::config {
     owner  => $::jenkins::user,
     group  => $::jenkins::group,
     mode   => '0755',
+    links  => follow,
   }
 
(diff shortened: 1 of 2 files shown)


hdoedens:
Sorry, I'll set-up travis first om my fork. When all tests succeed I'll do a pull request again.

PR #432, opened 2015-11-17
Symbolic link jobs folder

hdoedens:
Added plugins_base to be more flexible with custom plugin repositories. We have, for example, an artifactory repository as mirror for both jenkins.war files and plugins. The plugin files are in the path: http://servername/artifactory/jenkins-update/plugins. So no download part here.

Current implementation breaks the Jenkins installation when a symbolic link is used for the plugins or jobs folder. In such case the symbolic link is removed and replaced by an actual folder, effectively removing all jobs from Jenkins \o/.
I added a parameter to follow links in config for setup to enable links for jobs and plugins folder.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -19,6 +19,7 @@ class jenkins::config {
     owner  => $::jenkins::user,
     group  => $::jenkins::group,
     mode   => '0755',
+    links  => follow,
   }
 
   # ensure_resource is used to try to maintain backwards compatiblity with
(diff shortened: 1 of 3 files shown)


jhoblitt:
@hdoedens You feel that setting two params, `update_url` & `plugins_base`, is more convenient then using `source` directly?  I'm wondering if it wouldn't be better to have a single param on the `jenkins` class.

hdoedens:
@jhoblitt I don't like the 2 param construction, but I left it there to remain backwards compatible. Using the `source` param doesn't look clean to me. I would have to concatenate the download url outside the jenkins module. It feels a bit messy to me.
A single param on the Jenkins class would be a great solution.

ekohl:
Due to its age I'm closing this PR

PR #435, opened 2015-11-20
Fixed example for executors of jenkins master

DracoBlue:
Because they need to be configured in `jenkins::slave` and not `jenkins`


jhoblitt:
Hi @DracoBlue -- the example looks correct as-is.  The number of execution threads in the jenkins master is different than the number of threads on a swarm client.  I'm going to close this PR but please reopen it if I've misunderstood the intention.

PR #436, opened 2015-11-20
Use fqdn as fallback for `slave_name`

DracoBlue:
This is advertised in slave.pp, but does not work, yet.


jhoblitt:
@DracoBlue I thought that the swarm client defaulted to using the `fqdn` when `-name` isn't passed?

jhoblitt:
(if we do need to explicitly set the fqdn for the slave, it would be better do it in `slave.pp` rather than put conditional logic into the template)

ekohl:
If this is indeed needed, then a default in slave.pp would indeed be better. Due to its age I'm clsoing this now.

PR #437, opened 2015-11-30
Pass multiple --labels parameters

bigon:
Pass multiple --labels parameters instead of passing one with the label
space separated. This seems to be more resilient.

Change-Id: I990f22cbb7201f9269cda8f0ece6196a2c04434d

### DIFF
--- a/templates/jenkins-slave-defaults.erb
+++ b/templates/jenkins-slave-defaults.erb
@@ -70,9 +70,9 @@ if [ -n "$MASTER_URL" ]; then
   MASTER_URL_ARG="-master $MASTER_URL"
 fi
 
-if [ -n "$LABELS" ]; then
-  LABELS_ARG="-labels '$LABELS'"
-fi
+for LABEL in $LABELS; do
+  LABELS_ARG+="-labels $LABEL "
+done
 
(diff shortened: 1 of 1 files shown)


jhoblitt:
@bigon Could you be a little more explicit about the problem your seeing?

bigon:
We had some trouble to start the jenkins slave with space separated labels like "--labels foo bar baz"

With this patch we now have --labels foo --labels bar --labels baz and this is working

jhoblitt:
@bigon what version of the swarm client are you running?  I don't object to making this change but I'm surprised I haven't seen this issue myself.

bigon:
The jenkins slaves are Suse VM, I wrote I custom initscript for jenkins-slave (I should probably propose a PR too, see: https://github.com/bigon/puppet-jenkins/tree/vdab)

We use swarm-client 1.24

jhoblitt:
Could you test with swarm `2.0`?  We've run into at least one other case of param handling changing between swarm releases.

jhoblitt:
@bigon I am using multiple labels in production with swarm 2.0.  I'm going to close this PR but please reopen it if you can reproduce the problem.

PR #440, opened 2015-12-01
Only run jenkins create-job, if the job not already exists

buzzdeee:
Have the job xml file in hiera, and the job configured in hiera this way:
```
jenkins::job_hash:
  update_puppet_environment:
    config: "%{hiera('update_puppet_environment.xml')}"
```

deploying jenkins the first time, it creates the job, running jenkins again, the command fails, because the job already exists. Because of this, add an unless parameter, that checks if the job already exists.

### DIFF
--- a/manifests/job/present.pp
+++ b/manifests/job/present.pp
@@ -73,6 +73,7 @@ define jenkins::job::present(
   exec { "jenkins create-job ${jobname}":
     command => "${cat_config} | ${create_job}",
     creates => [$config_path, "${job_dir}/builds"],
+    unless  => "${jenkins_cli} get-job \"${jobname}\"",
     require => File[$tmp_config_path],
   }
 


buzzdeee:
problem observed with jenkins-1.609.3 on OpenBSD, see the other OpenBSD related PR ;)

jhoblitt:
Why do you think the `creates` param is failing to catch that the job already exists?

buzzdeee:
Doh, I had the localstatedir for OpenBSD case wrong, it was pointing to /var/jenkins, but actually should point to /var/jenkins/.jenkins because the plugins and jobs are subdirectories of that .jenkins directory. Fixed that in my other branch/PR, and removed that change from there again.

anyways, while looking into it, I found that there are all those neat native types, going to give those a try, since at least the user management with the defined types is not idempotent.

PR #441, opened 2015-12-03
add parameter proxy_host for jenkins::slave

TheMeier:
make it possible to use jenkins::slave if internet access is only possible via proxy


TheMeier:
#442

PR #456, opened 2015-12-21
Return nil for empty jenkins_plugins fact

mattkirby:
This commit updates the jenkins_plugins fact to return nil when there are no jenkins plugins. Without this change the fact generates a empty value and returns an empty string. When no jenkins plugins exist this causes the fact to show up in an unrelated context.

### DIFF
--- a/lib/puppet/jenkins/facts.rb
+++ b/lib/puppet/jenkins/facts.rb
@@ -33,7 +33,7 @@ module Puppet
           manifest = plugins[plugin]
           buffer << "#{plugin} #{manifest[:plugin_version]}"
         end
-        return buffer.join(', ')
+        return buffer.join(', ') unless buffer.length == 0
       end
     end
   end


rtyler:
Can you elaborate more on the "unrelated context" here? Since technically this does change the API, I'm curious more about why an empty string vs. `nil` is a problem

mattkirby:
By "unrelated context" I mean nodes that run puppet and are unrelated to jenkins. I encountered this when running `facter -p` on a node that does not use this module. It's a problem primarily from a user experience perspective since it appears odd to see the jenkins_plugins fact returned when jenkins does not run on the node.

bastelfreak:
Hi @mattkirby, are you still interested in getting this merged? If so, could you please rebase?

logicminds:
A better way to write this would be:  (not tested)

```
      # Return a list of plugins and their versions, e.g.:
      #   pam-auth 1.1, pmd 3.36, rake 1.7.8
      #
      # @return [String] Comma-separated version of "<plugin> <version>", empty
      #   string if there are no plugins
      def self.plugins_str
        plugins = Puppet::Jenkins::Plugins.available || {}
        plugins.keys.sort.map { |name|  "#{name} #{plugins[name][:plugin_version]}" }.join(", ") 
      end
```

ekohl:
I'm closing this for inactivity.

Issue #447, opened 2015-12-11
Systemd causes puppet idempotency issues


queeno:
Hi there

I've noticed this module is not idempotent when using it with systemd.

Every puppet run results in the following message being reported:

```
Notice: Compiled catalog for ci.vagrant in environment prod in 3.57 seconds
Notice: /Stage[main]/Jenkins::Service/Service[jenkins]/enable: enable changed 'false' to 'true'
Notice: Applied catalog in 8.52 seconds
```

I have reproduced it on CentOS 7.1.

jhoblitt:
@queeno What version of puppet are you running?

queeno:
I @jhoblitt 

The puppet version I am using is 4.2.3

I did some digging and turned out the issue occurs on systemd OSes (like RHEL/CentOS 7) only.

In order to determine whether a service is enable within systemd, [puppet issues](https://github.com/puppetlabs/puppet/blob/master/lib/puppet/provider/service/systemd.rb#L52) a `systemctl show --property LoadState --property UnitFileState --no-pager` command and then looks for the value of `UnitFileState`.

I found that - although systemd provides backward compatibility - `UnitFileState` doesn't get populated when using init.d scripts, inducing puppet to think that the service is disabled.

I can think of a couple of ways to fix this problem:

- Rectify the behaviour in puppet so it determines the state of the service in a different way (ie `systemctl is_enabled`)
- Use a systemd configuration file for Jenkins rather than init.d

cdenneen:
I found this bug (https://tickets.puppetlabs.com/browse/PUP-5353) which if you set provider on the service to redhat it will default back to SysV init.

I would propose the change to provider but since this module supports Ubuntu it would need to be constructed in a case statement or some other conditional.

jhoblitt:
@queeno I think that @cdenneen suggestion is the right approach here.  Depending on your perspective this is either a bug in systemd or a limitation in the systemd service provider, but I wouldn't hold my breath for this to get resolved upstream.

jhoblitt:
Anyone want to take a stab at a PR?

cdenneen:
#468 submitted

cdenneen:
#470 supersedes #468

PR #468, opened 2016-01-19
Update for issue #447

cdenneen:
Added conditional for fixing systemd issue

### DIFF
--- a/manifests/service.pp
+++ b/manifests/service.pp
@@ -6,11 +6,23 @@ class jenkins::service {
     fail("Use of private class ${name} by ${caller_module_name}")
   }
 
-  service { 'jenkins':
-    ensure     => $jenkins::service_ensure,
-    enable     => $jenkins::service_enable,
-    hasstatus  => true,
-    hasrestart => true,
+  if ($::operatingsystem != 'Amazon')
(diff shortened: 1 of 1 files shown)


manifests/service.pp
```diff
@@ -6,11 +6,23 @@
...
   }
 
-  service { 'jenkins':
-    ensure     => $jenkins::service_ensure,
-    enable     => $jenkins::service_enable,
-    hasstatus  => true,
-    hasrestart => true,
+  if ($::operatingsystem != 'Amazon')
+  and (($::operatingsystem != 'Fedora' and versioncmp($::operatingsystemrelease, '7.0') >= 0)
+  or  ($::operatingsystem == 'Fedora' and versioncmp($::operatingsystemrelease, '15') >= 0)) {
```
jhoblitt: My preference would be to put the platform conditional logic in `params.pp` and set the provider param via a variable rather than duplicate the service resource.  Does that make sense?

cdenneen: you'd have to specify the default provider for all types in this case... by doing this I avoided having to know the other providers and hardcode them

jhoblitt: I believe that `undef` should work for all other platforms.

PR #470, opened 2016-01-19
Updated for RedHat systemd systems to use redhat provider until PUP-5353 is fixed

cdenneen:
To address issue #447


jhoblitt:
@cdenneen This looks great.  I will merge it as soon as I have a chance to make some beaker runs (we really need to automate that...).

jhoblitt:
Oops -- wasn't quite ready to merge that but I'll leave it in.  Aww, the excitement of running master...

PR #472, opened 2016-01-20
Add folder plugin support for jobs

carroarmato0:
Allow creating jobs inside folders made by the Folder plugin (https://wiki.jenkins-ci.org/display/JENKINS/CloudBees+Folders+Plugin)

First, a Folder job needs to exist.

```
# Create Folder job
jenkins::job { 'MyFolder':
  config  => template('jenkins_jobs/folder.xml.erb'),
}
```

Next, we will create a regular job which we want to have added inside the folder we just defined:

```
# Create myproject-default job and add it to MyFolder folder
jenkins::job { "myproject-default":
  folder  => 'MyFolder',
  config  => template('jenkins_jobs/pipelines/tomcat/default.xml.erb'),
  require => Jenkins::Job['MyFolder'],
}
```

Signed-off-by: Christophe Vanlancker christophe.vanlancker@inuits.eu

### DIFF
--- a/manifests/job.pp
+++ b/manifests/job.pp
@@ -24,6 +24,9 @@
 #   ensure = 'present'
 #     choose 'absent' to ensure the job is removed
 #
+#   folder
+#     name of the folder where the job should be added to (using the Folder plugin)
+#
 #   difftool = '/usr/bin/diff -b-q'
 #     Provide a command to execute to compare Jenkins job files
 #
@@ -34,11 +37,16 @@ define jenkins::job(
   $jobname  = $title,
   $enabled  = 1,
   $ensure   = 'present',
+  $folder   = undef,
   $difftool = '/usr/bin/diff -b -q',
 ){
   include ::jenkins::cli
 
-  validate_string($difftool)
+  validate_string($difftool, $folder)
+
+  if $folder and (!defined(Jenkins::Plugin['cloudbees-folder'])) {
+    fail('cloudbees-folder plugin is required for the \'folder\' parameter.')
+  }
 
   Class['jenkins::cli'] ->
     Jenkins::Job[$title] ->
@@ -65,6 +73,7 @@ define jenkins::job(
       config   => $realconfig,
       jobname  => $jobname,
       enabled  => $enabled,
+      folder   => $folder,
       difftool => $difftool,
     }
   }
--- a/manifests/job/present.pp
+++ b/manifests/job/present.pp
@@ -13,24 +13,39 @@
 #   enabled = 1
 #     if the job should be enabled
 #
+#   folder = undef
+#     if the job should reside in a folder (using the Folder plugin)
+#
 define jenkins::job::present(
   $config,
   $jobname  = $title,
   $enabled  = 1,
+  $folder   = undef,
   $difftool = '/usr/bin/diff -b -q',
 ){
   include jenkins::cli
   include jenkins::cli::reload
 
-  validate_string($difftool)
+  validate_string($difftool, $folder)
 
   if $jenkins::service_ensure == 'stopped' or $jenkins::service_ensure == false {
     fail('Management of Jenkins jobs requires \$jenkins::service_ensure to be set to \'running\'')
   }
 
   $jenkins_cli        = $jenkins::cli::cmd
-  $tmp_config_path    = "/tmp/${jobname}-config.xml"
-  $job_dir            = "${::jenkins::job_dir}/${jobname}"
+
+  # Add support for Folders plugin
+  if $folder {
+    $realjobname     = "${folder}/${jobname}"
+    $job_dir         = "${::jenkins::job_dir}/${folder}/jobs/${jobname}"
+    $tmp_config_path = "/tmp/${folder}-${jobname}-config.xml"
+  }
+  else {
+    $realjobname     = $jobname
+    $job_dir         = "${::jenkins::job_dir}/${jobname}"
+    $tmp_config_path = "/tmp/${jobname}-config.xml"
+  }
+
   $config_path        = "${job_dir}/config.xml"
 
   # Bring variables from Class['::jenkins'] into local scope.
@@ -69,16 +84,16 @@ define jenkins::job::present(
 
   # Use Jenkins CLI to create the job
   $cat_config = "cat \"${tmp_config_path}\""
-  $create_job = "${jenkins_cli} create-job \"${jobname}\""
-  exec { "jenkins create-job ${jobname}":
+  $create_job = "${jenkins_cli} create-job \"${realjobname}\""
+  exec { "jenkins create-job ${realjobname}":
     command => "${cat_config} | ${create_job}",
     creates => [$config_path, "${job_dir}/builds"],
     require => File[$tmp_config_path],
   }
 
   # Use Jenkins CLI to update the job if it already exists
-  $update_job = "${jenkins_cli} update-job ${jobname}"
-  exec { "jenkins update-job ${jobname}":
+  $update_job = "${jenkins_cli} update-job ${realjobname}"
+  exec { "jenkins update-job ${realjobname}":
     command => "${cat_config} | ${update_job}",
     onlyif  => "test -e ${config_path}",
     unless  => "${difftool} ${config_path} ${tmp_config_path}",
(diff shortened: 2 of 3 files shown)


jhoblitt:
@rtyler / @matthewbarr - I'm not familiar with the folders plugin -- by any chance are either of you?

manifests/job.pp
```diff
@@ -34,6 +37,7 @@
   $jobname  = $title,
   $enabled  = 1,
   $ensure   = 'present',
+  $folder   = undef,
```
jhoblitt: Param validation is currently inconsistent in this module (#392) but I'm requesting that all new params have validation.

manifests/job/present.pp
```diff
@@ -13,10 +13,14 @@
...
 #     if the job should be enabled
 #
+#   folder = undef
+#     if the job should reside in a folder (using the Folder plugin)
+#
 define jenkins::job::present(
   $config,
   $jobname  = $title,
   $enabled  = 1,
+  $folder   = undef,
```
jhoblitt: Should be validated.

manifests/job/present.pp
```diff
@@ -29,8 +33,19 @@
...
 
   $jenkins_cli        = $jenkins::cli::cmd
-  $tmp_config_path    = "/tmp/${jobname}-config.xml"
-  $job_dir            = "${::jenkins::job_dir}/${jobname}"
+
+  # Add support for Folders plugin
+  if $folder {
+    $realjobname     = "${folder}/${jobname}"
+    $job_dir         = "${::jenkins::job_dir}/${folder}/jobs/${jobname}"
+    $tmp_config_path = "/tmp/${folder}/${jobname}-config.xml"
```
jhoblitt: Does it make sense to make `$folder` part of the temp file name instead of creating a subdir?  That would eliminate the need for an additional file resource.

manifests/job.pp
jhoblitt: If `$folder` is defined, it should be possible to add a dep along the lines of `Jenkins::Job[$folder] -> Jenkins::Job[$title]` so the user doesn't need to explicitly declare it.

carroarmato0: I added a link to the Folder plugin in the description.

In short:  the Folder Plugin allows you to organise jobs in "folders".

Traditionally you'd organise jobs in views, but if you have a lot of jobs and views, things will get very cluttered.

It's a Cloudbees supported plugin.

jhoblitt:
@carroarmato0 I'm not familiar with the folders plugin.  What is the failure mode if it is not present?  I'm guessing the sub-folders jobs are not visible to jenkins or does it break the master?

This PR needs rspec coverage and warrants a blurb in the README as it requires a non-default plugin.

manifests/job/present.pp
carroarmato0: The Folder plugin introduces the concept of namespaced jobs.
So stuff like:    MyProject/checkout    and   OtherProject/checkout  would be possible.

If I don't make $folder part of the temp file name you'd have duplicate   checkout-config.xml  files in /tmp.

Furthermore.  jobs who reside in a Folder have to be addressed with the folder name as it's part of the name space.

carroarmato0:
@jhoblitt jenkins will spit out exceptions in the log if the plugin is not loaded.

    Caused by: com.thoughtworks.xstream.mapper.CannotResolveClassException: com.cloudbees.hudson.plugins.folder.Folder

I could have a statement installing the plugin in case it's not present. But not sure if I had to handle the ability of a user wanting to pin the version of the plugin somewhere else.

What would be the best course of action for this?

manifests/job/present.pp
jhoblitt: My comment was on the path used for the *temp file*.  I was suggesting something along the lines of `$tmp_config_path = "/tmp/${folder}-${jobname}-config.xml"`

carroarmato0: Oh indeed, that'll work as well.

manifests/job.pp
jhoblitt: What I am asking for is `validate_string($folder)`.

manifests/job/present.pp
carroarmato0: I'll have to test and implement the absent state which is currently not handled.

jhoblitt:
@carroarmato0 I think all that's needed is a discussion in the README explaining that the plugin is required to use the folders feature and perhaps an example of the error message generated if it's not present.

carroarmato0:
@jhoblitt  I'm trying to make it optional by only declaring the plugin if you are actually using the  $folder  parameter and also checking if it hasn't been already declared somewhere else ( so that users may do some additional configurations for the plugin if wanted ).  Seems more user friendly :)

But struggling with a dependency loop atm.

carroarmato0:
Awesome, setting ensure => absent  already works with and without any modifications

manifests/job.pp
```diff
@@ -34,11 +37,16 @@
...
   $difftool = '/usr/bin/diff -b -q',
 ){
   include ::jenkins::cli
 
-  validate_string($difftool)
+  validate_string($difftool, $folder)
+
+  if $folder and (!defined(Jenkins::Plugin['cloudbees-folder'])) {
+    fail('cloudbees-folder plugin is required for the \'folder\' parameter.')
+  }
```
carroarmato0: @jhoblitt  having the plugin install automatically when someone uses the folder parameter without declaring it elsewhere causes a dependency loop which I didn't manage to solve.

So I think it's easier to just fail and inform the user to specifically add it themselves if they want to use the folder feature.

If they don't use it to begin with, everything works like before.

jhoblitt: @carroarmato0 I wasn't suggestion that the plugin be automatically installed -- I actually would prefer that not happen.

Note that `defined()` is parse order dependent AND it does not create an ordering relationship in the catalog.  If it's necessary to constrain in this way, it would be better to use the plugin fact but it would likely mean it would take multiple runs to to converge a catalog that both installs the plugin and a folder job.

What I was suggesting is that a dep between the sub-job and the parent folder be automatically declared instead of requiring the end-user to declare this relationship as you have done in your example manifest.

carroarmato0:
@jhoblitt   Added 2 rspec tests for the folder parameter:  with and without inclusion of the cloudbees-plugin,  some validation for $folder.

Anything else I still need to do in order for this to be merged? :)

carroarmato0:
Soo.. any more feedback needed on this?

jhoblitt:
@carroarmato0 There are still unaddressed comments on the PR, a blurb needs to be added to the README, and a rebase is needed.

jhoblitt:
Also, I'd really like to see rspec coverage of the path changes caused by putting the job in a subdir.

manifests/job.pp
```diff
@@ -34,11 +37,16 @@
...
   $ensure   = 'present',
+  $folder   = undef,
   $difftool = '/usr/bin/diff -b -q',
 ){
   include ::jenkins::cli
 
-  validate_string($difftool)
+  validate_string($difftool, $folder)
+
+  if $folder and (!defined(Jenkins::Plugin['cloudbees-folder'])) {
```
jhoblitt: It would be better to use the `jenkins_plugins` fact here:

    if $folder and (empty(grep([ $::jenkins_plugins ], 'cloudbees-folder'))) {

jhoblitt:
I took another look at this PR and I don't think the `folder` parameter is needed at all as this could be part of the resource name.  I tested using the folder plugin with the `jenkins_job` type and it worked as expected. Eg.

    jenkins_job { 'seeds/myfoo': ... }

bastelfreak:
Hi @carroarmato0, are you still interested in getting this merged? If so, could you please rebase against our latest master branch?

carroarmato0:
> Hi @carroarmato0, are you still interested in getting this merged? If so, could you please rebase against our latest master branch?

I haven't touched the code in 3 years.  Merging the code will likely be a pain. In the meantime at my company we switched to Groovy using the Jenkins Job DSL to handle to programmatically generate the jobs with folders.

So this PR is not useful, valuable to me. Can be dropped.

@jhoblitt mentioned in a previous comment that the parameter is not needed, which makes the PR also less relevant.

rnelson0:
@carroarmato0 Thanks for the feedback, we will close this now.

PR #473, opened 2016-01-20
validate all class/define params

### LINKED ISSUE #392: validate all DSL class/define params

jhoblitt:
Resolves #392

### DIFF
--- a/manifests/augeas.pp
+++ b/manifests/augeas.pp
@@ -42,35 +42,36 @@ define jenkins::augeas (
   $plugin           = false,
   $restart          = false,
 ) {
+  validate_string($config_filename)
+  if ! is_string($changes) and ! is_array($changes) {
+    fail('$changes must be string or array.')
+  }
+  if ! is_string($onlyif) and ! is_array($onlyif) {
+    fail('$onlyif must be string or array.')
+  }
+  validate_string($context)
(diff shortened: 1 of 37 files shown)


jhoblitt:
This is not yet ready for merge as I'm in the process of auditing.

jhoblitt:
This should be ready for merge now.  It is likely to break some manifests due to the new strict type checking but it should be trivial. Ie., removing quotes from around an integer value.

jhoblitt:
I'm going to go ahead and merge this so that it has time to circulate on master for a bit before another minor release (which I hope is coming soonish).

manifests/augeas.pp
```diff
@@ -42,35 +42,36 @@
...
+    true: {
       jenkins::plugin {$name:
         version       => $plugin_version,
         manage_config => false,
         before        => Augeas["jenkins::augeas: ${name}"],
       }
     }
-    false,'false': { # lint:ignore:quoted_booleans
-      #do nothing
+    false: {
```
PierreR: I am not sure why you want to keep the false case. As an outsider to the code, it looks idiotic ;-)

jhoblitt: Agreed.  Interested in a PR'ing a cleanup?

Issue #465, opened 2016-01-15
Core plugins won't upgrade


alexjfisher:
Hi

I decided to create jenkins::plugin resources for all the plugins in my installation (by manually updating all of them and then noting the versions).

Trouble is, on a fresh install, none of the core? plugins get updated. Each time puppet is run, it appears to download the plugin and then restart the service, but when I go to 'manage plugins' the plugin hasn't been upgraded to the version specified.

I've tried the latest release and the latest git version.

Here's the relevant bits of my profile.
```
class { '::jenkins':
    lts                => true,
    repo               => false,
    configure_firewall => false,
    config_hash        => $config_hash,
    localstatedir      => '/u01/jenkins',
    #plugin_hash        => $plugin_hash,
    require            => Rhn_channel['oraclelinux6-x86_64-jenkins'],
  }
contain '::jenkins'

jenkins::plugin { 'credentials':
    version => '1.24',
}
```

The installed version of 'credentials' remains at 1.18.  Hopefully some puppet debug log might help? :)
```
Debug: Exec[download archive credentials.hpi and check sum](provider=posix): Executing 'curl  -s -S -k -L -o /u01/jenkins/plugins/credentials.hpi 'https://updates.jenkins-ci.org/download/plugins/credentials/1.24/credentials.hpi''
Debug: Executing 'curl  -s -S -k -L -o /u01/jenkins/plugins/credentials.hpi 'https://updates.jenkins-ci.org/download/plugins/credentials/1.24/credentials.hpi''
Notice: /Stage[main]/Profile::Jenkins/Jenkins::Plugin[credentials]/Archive::Download[credentials.hpi]/Exec[download archive credentials.hpi and check sum]/returns: executed successfully
Debug: /Stage[main]/Profile::Jenkins/Jenkins::Plugin[credentials]/Archive::Download[credentials.hpi]/Exec[download archive credentials.hpi and check sum]: The container Archive::Download[credentials.hpi] will propagate my refresh event
Info: Archive::Download[credentials.hpi]: Scheduling refresh of Service[jenkins]
Debug: Archive::Download[credentials.hpi]: The container Jenkins::Plugin[credentials] will propagate my refresh event
Debug: /Stage[main]/Profile::Jenkins/Jenkins::Plugin[credentials]/File[/u01/jenkins/plugins/credentials.hpi.disabled]: Nothing to manage: no ensure and the resource doesn't exist
[...]

PR #475, opened 2016-01-21
Add ensure=>file to pinning file

alexjfisher:
Hi

This fix for https://github.com/jenkinsci/puppet-jenkins/issues/465 actually seems *too* simple.  It works for me, but I suspect I'm missing something! :)
But without an 'ensure' property, what did the file resource actually do?  Chown files if they happened to exist?

Kind Regards,
Alex

### DIFF
--- a/manifests/plugin.pp
+++ b/manifests/plugin.pp
@@ -103,6 +103,7 @@ define jenkins::plugin(
     }
 
     file { "${::jenkins::plugin_dir}/${plugin}.pinned":
+      ensure  => file,
       owner   => $::jenkins::user,
       group   => $::jenkins::group,
       require => Archive::Download[$plugin],
--- a/manifests/plugin.pp
+++ b/manifests/plugin.pp
@@ -26,6 +26,7 @@ define jenkins::plugin(
   $digest_string   = '',
   $digest_type     = 'sha1',
   $timeout         = 120,
+  $pin             = false,
   # deprecated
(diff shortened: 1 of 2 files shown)


jhoblitt:
@alexjfisher I agree it is odd that file resource doesn't have an `ensure` param.  I'm not sure if the design intent was to only set the permissions on the `.pinned` file, if it exists or to pin all managed plugins. Hmm.

sfhardman:
I think the trouble is this would pin all plugins, not just bundled ones.  Not sure how big of a deal that is.   It may only be an issue if a plugin is not bundled when puppet is applied, but gets added to the bundle on a later Jenkins release.  

I suspect the cleanest thing would be to add a "pinned" parameter and let people manually select whether to pin a plugin.  https://github.com/jenkinsci/puppet-jenkins/issues/380, and the result of the fix (https://github.com/jenkinsci/puppet-jenkins/commit/32c015e2c52a96ac08424314f42cc8f0f23d6e85) suggests that it is difficult to detect whether a plugin should be pinned automatically, and have it work on first and subsequent applies.

alexjfisher:
@sfhardman I've added a new parameter to control the behaviour as you suggested.  I actually called it 'pin' (instead of 'pinned').

sfhardman:
Works for me :-)

jhoblitt:
@rtyler Any thoughts on this?

PR #479, opened 2016-01-26
Introduced a helper to deal with augeas and configuration (and plugins)

vStone:


vStone:
Could somebody retry the build, I'm pretty sure it should pass :)

vStone:
Gonna fix some more tests and notifies

PR #480, opened 2016-01-26
An augeas helper define to deal with configs

vStone:
Helps with the configuration of a plugin and reloads (or restarts) jenkins after config changes.

### DIFF
--- a/spec/defines/jenkins_augeas_spec.rb
+++ b/spec/defines/jenkins_augeas_spec.rb
@@ -1,0 +1,220 @@
+require 'spec_helper'
+
+describe 'jenkins::augeas' do
+
+  let(:title) { 'myplug' }
+  let(:facts) {{ :osfamily => 'RedHat', :operatingsystem => 'CentOS' }}
+
+
+  #-------------------------------------------------------------------------------
+  #      |              o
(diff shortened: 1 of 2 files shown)


jhoblitt:
@vStone The quality of this code is great!  I need to find the time to think about this feature before merging it unless one of the other maintainers beats me to.

vStone:
Looks like this could use some more improvement. Now I require the cli class to be enabled. We could probably fall back to notifying the `Service` in stead of doing a reload/safe-restart when the cli is not used.

spec/defines/jenkins_augeas_spec.rb
```diff
@@ -0,0 +1,220 @@
...
+  describe "with plugin param wrong type" do
+    let (:params) {{:config_filename => 'foo.xml', :changes => [], :plugin => ['foo','bar'] }}
+    it do
+      is_expected.to raise_error(Puppet::Error, /is not a string/i)
+    end
+  end
+
+
+  #-------------------------------------------------------------------------------
+  #      |              o                             o
```
rtyler: this is silly :P

vStone: ¯\_(ツ)_/¯  its only a `git rm` away

PR #481, opened 2016-01-26
job_dir and plugin_dir defaulting from params but overrideable (Fixes #474)

### LINKED ISSUE #474: plugins_dir and job_dir don't default correctly
Hi

If you set ```localstatedir``` eg...
```
class { '::jenkins':
  localstatedir => '/data/jenkins'
}
```
I would not expect to also have to set ```plugins_dir``` and ```job_dir```.
But, if you don't, they still default to ```/usr/lib/jenkins/plugins``` and  ```/usr/lib/jenkins/jobs```.

vStone:
See #474: This fix allows setting the plugin and job dir in the params without
affecting the 'users' experience. I don't know why you would want to
override the job and plugin dir though. But this solution supports both
setting it in params and overriding it.


jhoblitt:
@vStone I'm not sure what the use case for this would be? AFAIK - there are no configuration values for these paths in jenkins -- a symlink is required as is mentioned in #403

vStone:
It was either this or revert putting the job_dir and plugin_dir in the params file. I could also cook that up

jhoblitt:
@vStone I suspect we should remove the `job_dir` & `plugin_dir` params completely from https://github.com/jenkinsci/puppet-jenkins/commit/00a90d4d162bca20eb1e04554def4ca45028affb  There hasn't been a stable release since I merged that PR so it should quasi-OK to remove them from the API.  What do you think?

vStone:
Get rid of them. People might be confused that they could adjust settings in jenkins to customize these

On January 26, 2016 7:15:02 PM CET, Joshua Hoblitt <notifications@github.com> wrote:
>@vStone I suspect we should remove the `job_dir` & `plugin_dir` params
>completely from
>https://github.com/jenkinsci/puppet-jenkins/commit/00a90d4d162bca20eb1e04554def4ca45028affb
>There hasn't been a stable release since I merged that PR so it should
>quasi-OK to remove them from the API.  What do you think?
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/jenkinsci/puppet-jenkins/pull/481#issuecomment-175151315

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

Issue #413, opened 2015-10-26
module should fail on unsupported platforms


jhoblitt:
This module should `fail` on any platform for which support is not expressed in `metadata.json.`

jhoblitt:
Possibility related to #410

PR #484, opened 2016-01-28
Bugfix/custom localdir with user: Fixes #462

### LINKED ISSUE #462: Create jenkins user before installing the package with manage_user
The usermod fails when using a custom localestatedir value due to jenkins already running after installation (debian-ish systems are 'cool' like that)

```
==> jenkins: Notice: /Stage[main]/Jenkins::Package/Package[jenkins]/ensure: ensure changed 'purged' to 'present'
==> jenkins: Error: Could not set home on user[jenkins]: Execution of '/usr/sbin/usermod -d /data/jenkins jenkins' returned 8: usermod: user jenkins is currently used by process 17664
==> jenkins: Error: /Stage[main]/Jenkins::Config/User[jenkins]/home: change from /var/lib/jenkins to /data/jenkins failed: Could not set home on user[jenkins]: Execution of '/usr/sbin/usermod -d /data/jenkins jenkins' returned 8: usermod: user jenkins is currently used by process 17664
```

vStone:
By splitting out the user creation, we can have a custom
localdir before the package (maybe) tries to create it or adjust it.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -14,46 +14,5 @@ class jenkins::config {
   )
   create_resources('jenkins::sysconfig', $config_hash)
 
-  $dir_params = {
-    ensure => directory,
-    owner  => $::jenkins::user,
-    group  => $::jenkins::group,
-    mode   => '0755',
-  }
 
-  # ensure_resource is used to try to maintain backwards compatiblity with
-  # manifests that were able to external declare resources due to the
-  # old conditional behavior of jenkins::plugin
-  if $::jenkins::manage_user {
-    ensure_resource('user', $::jenkins::user, {
-      ensure     => present,
-      gid        => $::jenkins::group,
-      home       => $::jenkins::localstatedir,
-      managehome => false,
-      system     => true,
-    })
-  }
-
-  if $::jenkins::manage_group {
-    ensure_resource('group', $::jenkins::group, {
-      ensure => present,
-      system => true,
-    })
-  }
-
-  $plugin_dir_params = $::jenkins::purge_plugins ? {
-    true    => $dir_params + {
-      'purge'   => true,
-      'recurse' => true,
-      'force'   => true,
-      'notify'  => Class['jenkins::service'],
-    },
-    default => $dir_params,
-  }
-
-  if $::jenkins::manage_datadirs {
-    ensure_resource('file', $::jenkins::localstatedir, $dir_params)
-    ensure_resource('file', $::jenkins::plugin_dir, $plugin_dir_params)
-    ensure_resource('file', $::jenkins::job_dir, $dir_params)
-  }
 }
--- a/manifests/user_setup.pp
+++ b/manifests/user_setup.pp
@@ -1,0 +1,60 @@
+# This class should be considered private
+#
+# It will optionally create the jenkins user and make sure all
+# directories have proper permissions setup.
+#
+# By having this in a separate class that is managed before
+# installing the package, we can effectivly override the
+# default local dir that is otherwise possibly created by
+# the package.
+#
+class jenkins::user_setup {
+
+  if $caller_module_name != $module_name {
+    fail("Use of private class ${name} by ${caller_module_name}")
+  }
+
+  $dir_params = {
+    ensure => directory,
+    owner  => $::jenkins::user,
+    group  => $::jenkins::group,
+    mode   => '0755',
+  }
+
+  # ensure_resource is used to try to maintain backwards compatiblity with
+  # manifests that were able to external declare resources due to the
+  # old conditional behavior of jenkins::plugin
+  if $::jenkins::manage_user {
+    ensure_resource('user', $::jenkins::user, {
+      ensure     => present,
+      gid        => $::jenkins::group,
+      home       => $::jenkins::localstatedir,
+      managehome => false,
+      system     => true,
+    })
+  }
+
+  if $::jenkins::manage_group {
+    ensure_resource('group', $::jenkins::group, {
+      ensure => present,
+      system => true,
+    })
+  }
+
+  $plugin_dir_params = $::jenkins::purge_plugins ? {
+    true    => $dir_params + {
+      'purge'   => true,
+      'recurse' => true,
+      'force'   => true,
+      'notify'  => Service['jenkins'],
+    },
+    default => $dir_params,
+  }
+
+  if $::jenkins::manage_datadirs {
+    ensure_resource('file', $::jenkins::localstatedir, $dir_params)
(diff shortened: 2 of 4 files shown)


vStone:
Looks like the tests fail due to the repo class having support for generic linux but the sysconfig define does not.

vStone:
@jhoblitt / @rtyler: Do we really want to assume that a unknown osfamily is a redhat variant? If we do, we should probably assume this for all `case $::osfamily` in the code...

jhoblitt:
Probably not.  The intent of #413 is to audit and cleanup platform handling.

jhoblitt:
I'm not enthusiastic about the name `pre_config`, as it is fairly ambiguous.  How about something like 'user_setup` or `role_account`?

manifests/config.pp
```diff
@@ -7,37 +7,10 @@
...
-  # old conditional behavior of jenkins::plugin
-  if $::jenkins::manage_user {
-    ensure_resource('user', $::jenkins::user, {
-      ensure     => present,
-      gid        => $::jenkins::group,
-      home       => $::jenkins::localstatedir,
-      managehome => false,
-      system     => true,
-    })
+  if ! has_key($::jenkins::config_hash, 'JENKINS_HOME') {
```
jhoblitt: If a `Jenkins::Sysconfig['JENKINS_HOME']` resource is already defined, this will fail.  I don't think it is safe to assume that it won't happen as `jenkins::sysconfig` is a public interface (I'd rather we didn't have both `config_hash` and the defined type but we are stuck for the moment). 

I think using `defined` to check for the existence of the resource might be marginally safer?  That doesn't prevent the end user from causing a duplicate resource declaration by having a `JENKINS_HOME` key in `config_hash` and a `jenkins::sysconfig` resource but I feel it should be more obvious to the end user that they shouldn't define the same value twice.

vStone:
I've renamed the class and in stead of only checking for either the resource in the hash or defined as a recourse, I do both.

vStone:
Rebased and attempt to fix some facter osfamily vs os['distro']['family']-ish issues. Needs a better solution in another pull request.

jhoblitt:
I'm afraid this isn't safe to merge because of the puppet 4.x breakage.

vStone:
I've put it in the same state as any other build. The beaker/docker builds fail, but everything else is green.

vStone:
@jhoblitt any remarks?

jhoblitt:
I've kicked travis.  I'm a bit nervous that all 3 acceptance tests are failing.

vStone:
```The command "bundle exec rake parallel_spec SPEC_OPTS='--format documentation --order random'" exited with 0.

Done. Your build exited with 1.
```
o.O

vStone:
Rebased on master.

vStone:
@jhoblitt Care to give this another set of eyes?

I rebased on master and all travis builds, except the one that should not be building (puppet 3.8.0 on ruby 2.2) are currently in the green.

manifests/config.pp
```diff
@@ -14,36 +14,13 @@
...
   create_resources('jenkins::sysconfig', $config_hash)
 
-  $dir_params = {
-    ensure => directory,
-    owner  => $::jenkins::user,
-    group  => $::jenkins::group,
-    mode   => '0755',
+  if (
+    ! defined(Jenkins::Sysconfig['JENKINS_HOME'])  and
+    ! has_key($::jenkins::config_hash, 'JENKINS_HOME')
```
jhoblitt: Is this necessary?  This is asking for parse order problems.

spec/classes/jenkins_user_setup_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+  let (:facts) { {
+    :osfamily => 'RedHat',
+    :operatingsystem => 'RedHat',
+    :operatingsystemrelease => '7.4',
+    :operatingsystemmajrelease => '7'
+  } }
+
+  context 'user_setup' do
+    context 'default' do
+      it { should contain_class('jenkins::config') }
```
jhoblitt: This seems unrelated to what is being tested?  It is a private class too.

spec/classes/jenkins_user_setup_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+      end
+    end
+    context 'unmanaged' do
+      let (:params) {{
+        :manage_user     => false,
+        :manage_group    => false,
+        :manage_datadirs => false,
+      }}
+      it { should_not contain_user('jenkins') }
+      it { should_not contain_user('jenkins') }
```
jhoblitt: Duplicate statement.

spec/classes/jenkins_user_setup_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+      it { should_not contain_group('jenkins') }
+      it { should_not contain_file('/var/lib/jenkins') }
```
jhoblitt: What about `.../{jobs,plugins}`?

manifests/config.pp
vStone: Not checking will make sure manifests fail when they have a JENKINS_HOME in the config hash or manually defined. I should probably document the use case for having a different localstatedir from the jenkins's user home dir that will require manage_user => false then.

vStone:
@jhoblitt: all remarks addressed.

vStone:
@jhoblitt It took a while, but: rebased.

vStone:
@jhoblitt request for re-review. Can we get this merged?

vStone:
@jhoblitt please either merge this or decline the pull request (with a better solution) ;)

vStone:
Now that we (voxpupuli) have ownership, can we re-discuss this @jhoblitt ? overriding the default location with a custom user is still a problem it seems. I'm open to alternative implementations to get this fixed. 

Note: a solution might be to disallow overriding the default /var/lib/jenkins location.

jhoblitt:
@vStone Could you rebase on current master?  This looks mergable as long as travis passes.

manifests/user_setup.pp
```diff
@@ -0,0 +1,53 @@
+# This class hould be considered private
+#
```
jhoblitt: It would be useful to have a blurb discussing why `user/group` management needs to be handled before the package is installed.

vStone: We had issues with using a custom home directory

jhoblitt: I meant a blurb as a comment on the class.

bastelfreak:
I restarted the failed acceptance test

vStone:
They seem to not want to work on travis. I did get a complete run on my machine  k(͡๏̯͡⊙)ゞ

PR #485, opened 2016-02-01
add ability to set java_args on slaves

adamcstephens:
Need for setting memory, etc on java VM for slaves.

### DIFF
--- a/manifests/slave.pp
+++ b/manifests/slave.pp
@@ -97,9 +97,11 @@ class jenkins::slave (
   $ensure                   = 'running',
   $enable                   = true,
   $source                   = undef,
+  $java_args                = '',
 ) inherits jenkins::params {
 
   validate_string($tool_locations)
+  validate_string($java_args)
 
   $client_jar = "swarm-client-${version}-jar-with-dependencies.jar"
   $client_url = $source ? {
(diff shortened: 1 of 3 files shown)


manifests/slave.pp
```diff
@@ -97,9 +97,11 @@
   $ensure                   = 'running',
   $enable                   = true,
   $source                   = undef,
+  $java_args                = '',
```
jhoblitt: The default could be `undef`

templates/jenkins-slave-defaults.erb
```diff
@@ -9,6 +9,7 @@ JAVA=/usr/bin/java
 # arguments to pass to java
 #JAVA_ARGS="-Xmx256m"
 #JAVA_ARGS="-Djava.net.preferIPv4Stack=true" # make jenkins listen on IPv4 address
+JAVA_ARGS="<%= @java_args -%>"
```
jhoblitt: It would be better to not declare the variable unless it is defined.  It is possible [although unlikely] that the user would want the value of `JAVA_ARGS` to be inherited from the env.

jhoblitt:
@adamcstephens This generally looks good.  I've made a few small comments.  Could you add comments to the .pp file and the README documenting the new param?

adamcstephens:
@jhoblitt I've updated based on your comments. There are no README docs, they reference the class file.

jhoblitt:
@adamcstephens I was asking for a blurb to added to the README but I'm going to merge this without waiting any longer.  Thank you for contributing!

PR #490, opened 2016-02-08
Remove extra include

PierreR:
I believe this extra check is a bad idea ... It makes the module rely on quite a specific puppet behaviour (and arguably a bad design).

Namely in Puppet this works

```
class { 'jenkins':}
include jenkins
```
but this won't:
``` 
include jenkins
class { 'jenkins':}
```
I would love to see the following include removed because I don't think it is terribly relevant to add such a check there.
That said, my main motivation is the fact that [language-puppet](https://github.com/bartavelle/language-puppet) doesn't currently support such idiom.

### DIFF
--- a/manifests/plugin.pp
+++ b/manifests/plugin.pp
@@ -32,7 +32,6 @@ define jenkins::plugin(
   $group           = undef,
   $create_user     = undef,
 ) {
-  include ::jenkins
 
   validate_bool($manage_config)
   validate_bool($enabled)


PierreR:
I quite don't understand why some tests are failing. I thought the `include` was merely a sanity check ... It is actually relied upon internally ?

jhoblitt:
@PierreR The inclusion is needed in order for the module to function without being dependent on parse order.

jhoblitt:
@PierreR The behavior of resource ordering in puppet has some suitability to it.  This this particular case, the `include` statement is ensuring that the `jenkins` class is evaluated before the current class, ensuring that it's variables are available in the manifest to be referenced.

See: https://docs.puppetlabs.com/puppet/4.3/reference/lang_classes.html#include-like-vs-resource-like

PierreR:
Thanks for taking the time to explain this.

PR #494, opened 2016-02-09
Add support for prefix configuration in the CLI config class.

danielpalstra:
This change adds a prefix configration to the jenkins::cli::config class. The prefix is then used in the CLI command to connect to a Jenkins servers which has a prefix configured (in the config_hash).

### DIFF
--- a/NATIVE_TYPES_AND_PROVIDERS.md
+++ b/NATIVE_TYPES_AND_PROVIDERS.md
@@ -31,7 +31,7 @@ However, the version included in this module has provides some additional
 setup.  The parameters used to override default values are:
 
 * `cli_jar`
-* `port`
+* `url`
 * `ssh_private_key`
 * `puppet_helper`
 * `cli_tries`
@@ -48,11 +48,19 @@ class { 'jenkins::cli::config':
 }
 ```
 
+An example of setting an alternative port number and an addition of a prefix.
+
+```
+class { 'jenkins::cli::config':
+  ssh_private_key => 'http://localhost:9999/awesome-jenkins',
+}
+```
+
 These values may also be set via facts with the same name after the prefix
 `jenkins_`.  Class parameters have precedence over fact values.
 
 * `jenkins_cli_jar`
-* `jenkins_port`
+* `jenkins_url`
 * `jenkins_ssh_private_key`
 * `jenkins_puppet_helper`
 * `jenkins_cli_tries`
--- a/NATIVE_TYPES_AND_PROVIDERS.md
+++ b/NATIVE_TYPES_AND_PROVIDERS.md
@@ -52,7 +52,7 @@ An example of setting an alternative port number and an addition of a prefix.
 
 ```
 class { 'jenkins::cli::config':
-  ssh_private_key => 'http://localhost:9999/awesome-jenkins',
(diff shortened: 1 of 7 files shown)


jhoblitt:
@danielpalstra Instead of interpolating fragments of the jenkins URL, I think it would be more flexible to be able to specify the entire URL.  What do you think?

danielpalstra:
@jhoblitt personally I like the setup that the "old" CLI uses. For example  https://github.com/jenkinsci/puppet-jenkins/blob/master/lib/puppet/parser/functions/jenkins_port.rb#L16 takes the port from the config_hash which I think is very clean since you only have to define the port once. Same applies for the prefix. I would consider this a long term solution for the cli config as a whole. 

Regarding interpolating the Jenkins URL. Can you think of any use cases where a different host is used than localhost? I also wouldn't expect that people connect to the CLI by using a https url. So interpolating the url feels fine by me. Can you share your thoughts on this one as well?

danielpalstra:
@jhoblitt what do you think of my proposal?

bjwschaap:
+1 
We need this as well. I like this particular solution since it is backwards-compatible and has very little impact. Interpolating the URL could be done in a separate enhancement request..

please merge

superseb:
+1

jhoblitt:
I still feel that removal of the `port` parameter completely and replacing it with a canonical URL would be a better solution.  Backwards compatibility is not a strong concern with this code, the API is documented as being unstable.

danielpalstra:
I assume you mean the `prefix` and `port` parameter. I will remove those and create a parameter that contains the complete url of the Jenkins node that the cli needs to connect to.

danielpalstra:
@jhoblitt as discussed I've removed the `port` param and replaced it with a canonical `url` param. This had impact on the docs and spec tests which I've updated as well.

NATIVE_TYPES_AND_PROVIDERS.md
```diff
@@ -48,11 +48,19 @@ class { 'jenkins::cli::config':
 }
 ```
 
+An example of setting an alternative port number and an addition of a prefix.
+
+```
+class { 'jenkins::cli::config':
+  ssh_private_key => 'http://localhost:9999/awesome-jenkins',
```
jhoblitt: That is unlikely to work...

jhoblitt:
@danielpalstra This looks good (other then the README typo)!  I'm going to hold on merging this until I have a change to try it but it should go in soon.

danielpalstra:
@jhoblitt sounds good. If you like I can send in a gist containing some sample code to test.

jhoblitt:
I've restarted travis.  It is starting to look like puppet 4.3 / ruby 1.9.3 may be a broken combo.

danielpalstra:
Jup, keeps on failing :)

jhoblitt:
A few of the other ruby 1.9.3 combos are also running for over 10 minutes.  Travis will kill a job if there is no output for 10 minutes.  I'm not sure why the ruby 1.9.3 runtimes are so much longer.

jhoblitt:
@danielpalstra The ruby 1.9.3/puppet 4.3 combo failed again and we are seeing this with other PRs as well.  I'm going to merge this.  Thank you for your patience!

PR #495, opened 2016-02-09
Use Active Directory realm as type

danielpalstra:
I adjusted the puppet_helper to configure Active Directory as a security realm.

### DIFF
--- a/files/puppet_helper.groovy
+++ b/files/puppet_helper.groovy
@@ -680,6 +680,21 @@ class Actions {
         ]
         break
 
+      //  active directory
+      case 'hudson.plugins.active_directory.ActiveDirectorySecurityRealm':
+        config = [
+          setSecurityRealm: [
+            (className): [
+              realm.domain,
+              realm.site,
+              realm.bindName,
+              realm.bindPassword,
+              realm.groupLookupStrategy,
+            ],
+          ],
+        ]
+        break
+
       // constructor with no arguments
       // "Delegate to servlet container"
       case 'hudson.security.LegacySecurityRealm':


jhoblitt:
:+1: This looks good.  I will merge it as soon as travis completes (although, sadly, there is no unit test coverage of the groovy code).

danielpalstra:
@jhoblitt any thoughts on implementing unit tests for groovy? I have some sample code for usage with an AD.

jhoblitt:
@danielpalstra I think the best [behavioral] path would be to add acceptance tests that exercise the natives types which interact with the groovy code.

danielpalstra:
@jhoblitt :+1:  acceptance tests sound like a plan. With the addition of the hudson.security.AuthorizationStrategy$Unsecured jenkins_authorization_strategy an Active Directory back-end shouldn't be necessary. I'll open an issue as soon as I have the time.

jhoblitt:
Fabulous.  Thank you!

files/puppet_helper.groovy
```diff
@@ -680,6 +680,21 @@ class Actions {
         ]
         break
 
+      //  active directory
+      case 'hudson.plugins.active_directory.ActiveDirectorySecurityRealm':
+        config = [
+          setSecurityRealm: [
+            (className): [
```
cdenneen: @danielpalstra @jhoblitt If I'm reading this correctly the arguments here need to be updated? @danielpalstra are you using latest version of the active_directory plugin?
I need to set site as blank and there should be a realm.server parameter?

PR #497, opened 2016-02-14
Conditionally include apt class

ernetas:
Because some other Puppet module might have done this earlier.


rtyler:
To my knowledge there isn't an issue with redundant includes (as opposed to `class` resource declarations).

What is the error you're seeing that prompted this change?

ernetas:
Oh, right. Good point. 

I had `class { 'apt': }` used in one module. Going to patch that one instead then.

PR #500, opened 2016-02-16
Added seed_only Option to Jobs for Bootstrapping with JobDSL Plugin

elconas:
Normally the Jenkins manifest creates jobs and tries to do its best (XML mangling) to be idempotent. 

Sometimes you want to bootstrap Jenkins another way - e.g. with JobDSL Plugin. 

For this we just need to create a seed job, which then builds all jobs using job dsl. In this case Puppet should not care about the job settings in Jenkins anymore (also good for trainings). 

The seed_only parameter allows for this.


rtyler:
This looks reasonable but lacks some [rspec-puppet](http://rspec-puppet.com) test cases to go along with it to make sure that the `jenkins::job::present` logic is covered properly.

If you wish to tackle that, that would be great, otherwise I'll put this in my queue of pull requests to prepare unit tests for :)

elconas:
I will give it a try next week.

jhoblitt:
Instead of naming the parameter `seed_only`, how about `replace` since this is an well known semantic from the file resource? Other suggestion: `update` or `create_only`.  I'm concerned that `seed` in this context will not be intuitive for end-users.

vStone:
Both the define (jenkins::job/jenkins::job::present) and the native jenkins_job type support the `replace` parameter now. This can probably be closed. @elconas, your thoughts?

PR #503, opened 2016-02-24
Adds Beaker docker testing to .travis.yml

petems:
* Adds CentOS 6 Beaker docker file
* Adds travis matrix entry to run beaker-docker tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,5 +15,15 @@ env:
   - PUPPET_VERSION="~> 4.1.0"
   - PUPPET_VERSION="~> 4.2.0"
   - PUPPET_VERSION="~> 4.3.0"
+matrix:
+  include:
+    - rvm: '2.1'
+      sudo: required
+      services: docker
+      env: PUPPET_VERSION="3.8.4" BEAKER_set="centos-6-docker"
+      script: bundle exec rake acceptance
+      bundler_args: --without development
+  allow_failures:
+    - env: PUPPET_VERSION="3.8.4" BEAKER_set="centos-6-docker"
 notifications:
   email: false
--- a/Rakefile
+++ b/Rakefile
@@ -22,6 +22,11 @@ PuppetLint::RakeTask.new(:lint) do |l|
   l.log_format = "%{path}:%{linenumber}:%{check}:%{KIND}:%{message}"
 end
 
+desc "Run acceptance tests"
+RSpec::Core::RakeTask.new(:acceptance) do |t|
+  t.pattern = 'spec/acceptance'
+end
+
 PuppetSyntax.exclude_paths = exclude_paths
 
 task :default => [:lint, :validate, :spec]
--- a/spec/acceptance/nodesets/centos-6-docker.yml
+++ b/spec/acceptance/nodesets/centos-6-docker.yml
@@ -1,0 +1,12 @@
+HOSTS:
+  centos-6-x64:
+    platform: el-6-x86_64
+    hypervisor : docker
+    image: centos:6
+    docker_preserve_image: true
+    docker_cmd: '["/sbin/init"]'
+    docker_image_commands:
+      - 'yum install -y crontabs tar wget'
+CONFIG:
+  type: foss
+  log_level: debug


jhoblitt:
Is beaker officially compatible with ruby 2.2 now?

petems:
I believe has for a while according to https://tickets.puppetlabs.com/browse/BKR-537

petems:
It looks like the issue is with beaker-rspec and some versions rspec, this seems to be the workaround: https://github.com/puppetlabs/puppetlabs-firewall/pull/597/files

petems:
To simplify things I switched it back to 2.1.0 anyways

jhoblitt:
Awww, the ticket likely explains the failure I've seen.  This module isn't pinning the rspec version so it should, in theory, be able to work.

jhoblitt:
Tangentially, https://github.com/jenkinsci/puppet-jenkins/pull/473 fixes the unit tests under puppet 4.2+.

petems:
Cool, was gonna say, those tests take ages and 1.9.3 is EOL now :smile:

jhoblitt:
I thought #473 has already been merged. so I just put it in.  It probably causes a merge conflict in the `.travis.yaml`.  My bad. ;(

1.9.3 is in the matrix as this is still the low bar for puppet 4.

jhoblitt:
It looks like travis lost a worker mid-build.  I have restarted the job.

.travis.yml
```diff
@@ -15,5 +15,15 @@ env:
   - PUPPET_VERSION="~> 4.1.0"
   - PUPPET_VERSION="~> 4.2.0"
   - PUPPET_VERSION="~> 4.3.0"
+matrix:
+  include:
+    - rvm: '2.1'
+      sudo: required
+      services: docker
+      env: PUPPET_VERSION="3.8.4" BEAKER_set="centos-6-docker"
+      script: bundle exec rspec spec/acceptance/class_spec.rb
```
jhoblitt: Why not run all of (our vast collection...) of acceptance tests?

petems: I was mainly worried about the time it would take, but I can change it to run everything if required? :smile:

petems: Fixed :+1:

petems:
Travis died mid build again :/

jhoblitt:
I've kicked it again.  If this keeps up, we may need to stand up a jenkins instance ;)

jhoblitt:
And I've kicked it again... what is going on?

rtyler:
@jhoblitt I'm not against spinning this up in the [ci.jenkins-ci.org](https://ci.jenkins-ci.org) instance since it is fairly "official"

Should I file a ticket for that?

jhoblitt:
@rtyler That certainly wouldn't hurt!  From looking at the other jobs, it appears that the GitHub Pull Request Builder plugin is already installed.

petems:
@jhoblitt Does it make sense to remove the 1.9.3 tests if they're causing the issues? https://github.com/jenkinsci/puppet-jenkins/issues/509

jhoblitt:
@petems I suspect we may have to but note that puppet 4.x is technically supported on 1.9.3 and we are not using the AIO packages with travis.

petems:
@jhoblitt Can we move the 1.9.3 tests into allow_failure?

rtyler:
@petems 1.9.3 is still the default Ruby packaged up for a lot of people, myself included, running not-the-latest-versions of PE.

The error with the build in Travis isn't a segfault but something stopped outputting anything which to me means that somebody went either into an infinite loop or prompted for something on `stdin`.

I do not wish to remove 1.9.3. testing plz

petems:
Ok, will kill the issue then! :+1:

jhoblitt:
It looks like the ruby 1.9.3 build times have pretty consistently been longer than ruby 2.1.0.  Here is a build from ~4 months ago. 

https://travis-ci.org/jenkinsci/puppet-jenkins/builds/86419131

Puppet 4.3.0 is just a tad slower than previous puppet versions, which is why we are all of a sudden hitting the travis 10min output limit.  However, at least once I have seen other ruby 1.9.3 combos fail as well.  We need to identify what test is taking slow long to execute.  As an interm, I think it would be reasonable to exclude the ruby 1.9.3/puppet 4.3 combo from the matrix.

Since I have convinced myself that the failure is not related to anything in this PR (it is also happening in #508), I'm going to go ahead and merge this.

@petems - I really appreciate the time you have put into this PR.  This module is in desperate need of acceptance tests that cover the jenkins<->puppet interface that rspec-puppet can't address.

PR #518, opened 2016-03-03
Plugins from updatecenter performance fixes

### LINKED ISSUE #517: slow unit test causing travis failures
y so slow!

This cause of this slow unit test (particularly under ruby 1.9.3) needs to be hunted down and destroyed.

https://travis-ci.org/jenkinsci/puppet-jenkins/jobs/113456469#L1721-L1722

petems:
Fairly extreme, I can tone this down if needed? But it does fix #517

* The test is mainly to check the method works with a given json file
* In Ruby >= 2.0 it only takes 15 seconds to run, in 1.9.3 it can be 2 minutes or greater causing travis issues
* Changes contexts to explicitly state what the test is checking
* Actually check that hash has correct values

I can change it to completely just use 1 plugin and remove the if else loop around 1.9.3. The purpose of the test is to check that the method can parse a given JSON, the number is kinda inconsiquential. Ruby 1.9.3 has the json library built in anyway, so it's NBD right?


jhoblitt:
@petems I don't think this is at all unreasonable.  Travis hasn't even registered the PR yet but I think should go in as long as travis passes.

spec/unit/jenkins_plugins_spec.rb
```diff
@@ -137,20 +137,34 @@
 
   describe '.plugins_from_updatecenter' do
     subject(:plugins) { described_class.plugins_from_updatecenter(fixture) }
-    let(:fixture) { File.expand_path(File.dirname(__FILE__) + '/../fixtures/update-center.json') }
 
-    it { should be_instance_of Hash }
-    # Our fixture file currently has 870 plugins in it
-    its(:size) { should eql 870 }
+    if RUBY_VERSION == '1.9.3'
```
jhoblitt: It would be nice to add a comment here about why this was done.  Over time, re-factoring often makes it hard to track don't commit messages with just `git blame`.

petems: Cool, I generally keep the info in git commits because comments can sometimes be a smell, but I've added it to both to cover the bases :+1:

jhoblitt: :+1:

jhoblitt:
@petems :+1:

PR #523, opened 2016-03-04
Dependency correction when manage_slave_user is false

james-powis:
Corrected Slave Dependency with when manage_slave_user is false and manage_client_jar is true

### DIFF
--- a/manifests/slave.pp
+++ b/manifests/slave.pp
@@ -230,15 +230,27 @@ class jenkins::slave (
 
   #a Add jenkins slave user if necessary.
   if $manage_slave_user {
-    user { 'jenkins-slave_user':
-      ensure     => present,
-      name       => $slave_user,
-      comment    => 'Jenkins Slave user',
-      home       => $slave_home,
-      managehome => $manage_user_home,
-      system     => true,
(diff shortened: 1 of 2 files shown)


jhoblitt:
So now that I actually think about it, this will have essentialy the same problem... I think what we need to do is have a conditional that declares the relationship when both  `$manage_slave_user  and if `$manage_client_jar` are true.  What do you think?

james-powis:
the way I have gotten around this in the past has been to conditionally declare the same resource twice:
`if $manage_slave_user
  if $manage_client_jar
    user {'jenkins-slave_user':
      ...
      before => Archive['get_swarm_client']
    }
  } else {
    user {'jenkins-slave_user':
      ...
    }
}
`
What are your thoughts on that?

jhoblitt:
We can express this without having to duplicate the resource.  Something along the lines of (untested):

```
  if ($manage_slave_user and $manage_client_jar) {
    User['jenkins-slave_user']->
      Archive['get_swarm_client']
  }
```

jhoblitt:
:+1: Looks good.  Thank you contributing!

PR #504, opened 2016-02-25
Make user creation and deletion idempotent

mmickan:
Resolves #454

PR #524, opened 2016-03-06
Check password to see if it's changed when adding/updating credentials.

mterzo:
If the password has changed run the update/create credential command.

### DIFF
--- a/manifests/credentials.pp
+++ b/manifests/credentials.pp
@@ -50,7 +50,7 @@ define jenkins::credentials (
           "'${description}'",
           "'${private_key_or_path}'",
         ],
-        unless  => "\$HELPER_CMD credential_info ${title} | grep ${title}",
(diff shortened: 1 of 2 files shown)


jhoblitt:
The implications of this is it would update the password string but no other values?

mterzo:
Which is the current behavior.

jhoblitt:
The current behavior is, I believe, to only create the user if it doesn't exist.  I'm concerned it will be confusing if only one of the user values is updated.  The new `jenkins_user` type handles this correctly.  https://github.com/jenkinsci/puppet-jenkins/blob/master/NATIVE_TYPES_AND_PROVIDERS.md#jenkins_user

jhoblitt:
This PR is essentially a subset of #504 .

bastelfreak:
Hi @mterzo, are you still interested in getting this merged? If so, could you please rebase against our latest master branch?

PR #527, opened 2016-03-07
make centos-6-docker acceptance tests required

jhoblitt:


jhoblitt:
@petems Just FYI - I broke the centos-6-docker with my PR that replaced `camptocamp/archive` with `puppet/archive`.  It's pretty clear that we need to make that test configuration mandatory.

PR #531, opened 2016-03-08
add StringCredentialsImpl support to jenkins_credentials

### LINKED ISSUE #373: Credential types
I'm trying to configure the GitHub plugin using Puppet as part of that I need to provide a credential for the API token which requires a 'secret text' credential, currently this is not possible using this module.

Would it not be better to add a defined type for each kind of credential rather than one cluttered type?

Also `jenkins::credentials` really should be `jenkins::credential`, seems there is a PR to address this #311.

jhoblitt:
resolves #373


jhoblitt:
based on top of #530

jhoblitt:
rebased on master after the merge of #530

PR #540, opened 2016-03-29
make jenkins_job type cloudbees-folder aware

jhoblitt:
If the resource name contains slashes, treat them as a
`cloudbees-folder` path and autorequire all parent folders.

Eg., `foo/bar/myjob` would autorequire the `jenkins_job` resources named
`foo` and `foo/bar`.

relates to #472


rtyler:
Merge when travis goes green, LGTM

jhoblitt:
@rtyler Great!

PR #550, opened 2016-04-04
replace centos-7-docker fakesystemd with classic flavor

jhoblitt:


jhoblitt:
@petems Would you be willing to take a look at this PR?  I found the `dockerfile` option is the beaker source but I'm unsure if this is the best practice approach.

petems:
@jhoblitt Looks good to me! Honestly, I'm not the best person to ask :sweat_smile: I mainly stood on the shoulders of giants and re-used a lot of the code by people like @electrical

I guess having a Dockerfile is easier to maintain because you can use the standard Docker plumbing (linting, running outside of tests, even uploading a Beaker docker version to a registry), rather than having to parse a yml file.

What's the benefit of the entrypoint being systemd itself rather than init?

jhoblitt:
@petems I believe (?) that the only other way to setup the `MOUNT` needed by systemd would be to publish a image on docker hub, which felt like it would make changes more complicated than opening a PR.

I will go ahead and merge this but am willing to revise it if a better method is illustrated.

PR #552, opened 2016-04-06
rubocop

jhoblitt:


jhoblitt:
@jenkinsci/puppet-jenkins-developers How do people feel about requiring single quotes for literal strings?  I'm mildly in favor of it but this point of style seems to be pretty inconsistent between ruby code-bases.  The cop currently finds 253 offenses in this module.

rtyler:
@jhoblitt I'm very much in favor of using single quotes for literal strings, such that double-quotes are used only for strings which are intended to be interpolated.

PR #555, opened 2016-04-07
Abstract variable to expose for parameterisation of plugins source

chattr:
This is needed so we can continue to use your module alongside our existing infrastructure. Let me know if you have any comments or suggestions. Thanks.

### DIFF
--- a/manifests/master.pp
+++ b/manifests/master.pp
@@ -2,11 +2,14 @@
 #
 #
 class jenkins::master (
-  $version = $jenkins::params::swarm_version
+  $version = $jenkins::params::swarm_version,
+  $source  = undef,
 ) inherits jenkins::params {
   validate_string($version)
+  validate_string($source)
 
   jenkins::plugin {'swarm':
-    version => $version ,
+    version => $version,
+    source  => $source,
   }
 }


rtyler:
This doesn't look problematic to me, but it is always appreciated to have RSpec Puppet examples introduces with new parameters/features

chattr:
Thanks for taking a look so soon. I've had a dig around and found some part of what I think you need:

```
$ cat jenkins_master_spec.rb
...
let(:params) { { :source => 'https://updates.jenkins-ci.org/download/plugins/swarm/2.0/swarm.hpi' } }
...
```
That looks incomplete to me; please can you suggest how a complete test might look?

As you've probably guessed by now, I'm not a ruby dev and haven't written ruby tests before, so please bear with me.

jhoblitt:
I don't have a strong objection to this but I really don't want to see `jenkins::master` passing through all of `jenkins::plugin`'s parameters.  It isn't very difficult for the end user to replace `include ::jenkins::master` in the manifest with ` jenkins::plugin { 'swarm': ... }`.

bastelfreak:
Hello. I'm going to close this PR due to inactivity. Please reopen and rebase it if you're still interested in it.

PR #564, opened 2016-04-18
5th parameter is server list for ActiveDirectory

cdenneen:
Looking at this the 5th parameter is the server string not the GroupLookupStrategy.
The arguments only seem to be taking 5 parameters even though it seems from the init (https://github.com/jenkinsci/active-directory-plugin/blob/8a86022e18cb81dd4bc0e7ae9afd5a8cf048d894/src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java#L146) that it might be able to take 6 (https://github.com/jenkinsci/active-directory-plugin/blob/8a86022e18cb81dd4bc0e7ae9afd5a8cf048d894/src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java#L146-L150).

### DIFF
--- a/files/puppet_helper.groovy
+++ b/files/puppet_helper.groovy
@@ -730,7 +730,7 @@ class Actions {
               realm.site,
               realm.bindName,
               realm.bindPassword,
-              realm.groupLookupStrategy,
+              realm.server,
             ],
           ],
         ]


jhoblitt:
I've kicked the failed travis builds.

jhoblitt:
The 1.43 tag has both the 5 and 6 argument constructors:

https://github.com/jenkinsci/active-directory-plugin/blob/active-directory-1.43/src/main/java/hudson/plugins/active_directory/ActiveDirectorySecurityRealm.java#L150

What version are you testing with?  Will this plugin work with a regular LDAP server?  If so, it should be possible to write acceptance tests to make sure it doesn't break.

cdenneen:
1.43 is what I'm using.  I see the 6th argument on line 150 but if you pass it, it blows up.
I'm not sure if this works with generic LDAP or not.

jhoblitt:
There is no test coverage of this and I'm not a user of the AD plugin.  I'm going to go ahead and merge this since it shouldn't breake anything unrelated.

cdenneen:
@jhoblitt so the `groupLookupStrategy` parameter doesn't work, even with this.
Adding the additional parameter ends up in a groovy error guessing because it's too many parameters. Any idea of how to get the additional parameters to pass since the activedirectory plugin referenced above appears to accept them.

PR #568, opened 2016-04-26
Added systemd support

ifireball:
Systemd allows for auto-restarting the client and generally works better on EL >= 7

### DIFF
--- a/files/jenkins-slave.service
+++ b/files/jenkins-slave.service
@@ -1,0 +1,16 @@
+[Unit]
+Description=Jenkins slave service
+Requires=network.target
+After=network.target
+
+[Service]
+Type=simple
+ExecStart=/usr/local/sbin/jenkins-slave
+Restart=always
+RestartSec=60
+StartLimitInterval=0
+SuccessExitStatus=143
+
+[Install]
+WantedBy=multi-user.target
+
--- a/files/jenkins-slave.sh
+++ b/files/jenkins-slave.sh
@@ -1,0 +1,15 @@
+#!/bin/sh
+JENKINS_CONFIG=/etc/sysconfig/jenkins-slave
+
+# Read config
+[ -f "$JENKINS_CONFIG" ] && . "$JENKINS_CONFIG"
+
+if [ -x /sbin/runuser ] ; then
+    RUNUSER=runuser
+else
+    RUNUSER=su
+fi
+
+# Run swarm client
+exec $RUNUSER - $JENKINS_SLAVE_USER \
+    -c "$JAVA $JAVA_ARGS -jar $JENKINS_SLAVE_JAR $JENKINS_SLAVE_ARGS"
--- a/manifests/slave.pp
+++ b/manifests/slave.pp
@@ -1,30 +1,37 @@
 # == Class: jenkins::slave
 #
-# This module setups up a swarm client for a jenkins server.  It requires the swarm plugin on the Jenkins master.
+# This module setups up a swarm client for a jenkins server.  It requires the
+# swarm plugin on the Jenkins master.
 #
 # https://wiki.jenkins-ci.org/display/JENKINS/Swarm+Plugin
 #
-# It allows users to add more workers to Jenkins without having to specifically add them on the Jenkins master.
+# It allows users to add more workers to Jenkins without having to specifically
+# add them on the Jenkins master.
 #
 # === Parameters
 #
 # [*slave_name*]
-#   Specify the name of the slave.  Not required, by default it will use the fqdn.
+#   Specify the name of the slave.  Not required, by default it will use the
+#   fqdn.
 #
 # [*masterurl*]
-#   Specify the URL of the master server.  Not required, the plugin will do a UDP autodiscovery. If specified, the autodiscovery will be skipped.
+#   Specify the URL of the master server.  Not required, the plugin will do a
+#   UDP autodiscovery. If specified, the autodiscovery will be skipped.
 #
 # [*autodiscoveryaddress*]
 #   Use this addresss for udp-based auto-discovery (default: 255.255.255.255)
 #
 # [*ui_user*] & [*ui_pass*]
-#   User name & password for the Jenkins UI.  Not required, but may be ncessary for your config, depending on your security model.
+#   User name & password for the Jenkins UI.  Not required, but may be ncessary
+#   for your config, depending on your security model.
 #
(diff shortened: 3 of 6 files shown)


files/jenkins-slave.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
+JENKINS_CONFIG=/etc/sysconfig/jenkins-slave
+
+# Read config
+[ -f "$JENKINS_CONFIG" ] && . "$JENKINS_CONFIG"
+
+if [ -x /sbin/runuser ] ; then
+    RUNUSER=runuser
+else
+    RUNUSER=su
```
jhoblitt: I don't think this wrapper script is needed at all since the user can be declared in the systemd service unit?

manifests/slave.pp
```diff
@@ -174,13 +174,38 @@
...
-        ensure => 'file',
-        mode   => '0755',
-        owner  => 'root',
-        group  => 'root',
-        source => "puppet:///modules/${module_name}/jenkins-slave.${::osfamily}",
-        notify => Service['jenkins-slave'],
+      $systemd_min_release = $::osfamily ? {
+        'RedHat' => $::operatingsystem ? {
+          'Fedora' => 21,
+          default  => 7,
```
jhoblitt: Note that Debian Jessie and Ubuntu 16.04 also have systemd but I am fine with starting with just RH variants.

manifests/slave.pp
```diff
@@ -174,13 +174,38 @@
...
+        },
+        default => undef,
+      }
+      if $systemd_min_release {
+        $have_systemd = $::operatingsystemmajrelease >= $systemd_min_release
```
jhoblitt: I know this modue has a systemd fact: https://github.com/camptocamp/puppet-systemd/blob/master/lib/facter/systemd.rb, perhaps that would be easier then having to whitelist each platform?

files/jenkins-slave.sh
ifireball: The wrapper was created so settings can still be stored in the defaults/sysconfig file and used from it.

manifests/slave.pp
ifireball: I can't test on those atm.

manifests/slave.pp
ifireball: If adding such a dependency is acceptable, I'll try to add it, it may make unit testing harder though

adamcstephens:
Can/should we purge the init script when deploying the systemd unit file?

ifireball:
Since I've only been running this on fresh-installed hosts, and the older sys-v based code simply did not work for me, I don't have an environment to test purging, so I'd rather not write something I cannot test.

files/jenkins-slave.sh
jhoblitt: Why doesn't `EnvironmentFile` work?

manifests/slave.pp
jhoblitt: I would be OK with adding a dep on that module.  I know if it would or wouldn't be a better approach, just pointing it out as an implementation option.

jhoblitt:
@adamcstephens raises a valid point.  We probably don't want to install a sysv script when usinig a service file.

files/jenkins-slave.sh
adamcstephens: EnvironmentFile doesn't pass the file through a shell interpreter in my understanding, its just for key=value. The logic in the sysconfig file thus wouldn't work.

https://www.freedesktop.org/software/systemd/man/systemd.exec.html#EnvironmentFile=

jhoblitt: Aw, good point.  I was thinking of the jenkins master sysconfig which does have the conditional logic.  This seems reasonable then.  I would suggest relocating the `jenkins-slave` script to be under `$libdir` since there is no guarantee that `/usr/local/bin` exists.

ifireball:
The code does not install a sysv script along side the service file, it just doesn't remove it if it was placed there

jhoblitt:
@ifireball I believe @adamcstephens request was to actively `ensure => absent` the sysv init script when the systemd service file is in use.  This would be to prevent users from end up with both a system service and sysv init script when updating this module.

adamcstephens:
Correct. For users who are migrating from the existing code we will want to remove the sysvinit. That may strand the existing running process, but is better than trying to start two processes on boot.

ifireball:
CI failures do not seem related to changes in this patch.
Please help.

ifireball:
Now manually verified to work well both on EL6 (e.g. work the old way) and also on EL7 (be systemd compatible)

ifireball:
Ping?

jhoblitt:
travis is failing due to linter warnings that need to be fixed.  I'd also like to see the wrapper script moved out of `/usr/local` per previous comments.

ifireball:
Now I sees to only have the same fails as master in travis.

I couldn't find $libdir mentioned in puppet docs outside the context of the conf file (see [1]) so I just moved the script file to '/usr/sbin/' for now. I hope that is standard enough.

[1]: https://docs.puppet.com/puppet/latest/reference/lang_facts_and_builtin_vars.html#special-variables-added-by-puppet

jhoblitt:
The acceptance tests are broken again under travis because of jenkins 2.x.  :( There is a PR in progress add 2.x support.

By `$libdir`, I was referring to the top level parameter `$::jenkins::libdir`: https://github.com/jenkinsci/puppet-jenkins/blob/master/manifests/init.pp#L197

ifireball:
Hosts with 'jenkins::slave' would not necessarily include the 'jenkins' class. It seems to me unsafe to assume '/usr/lib/jenkins' would exist. (Random check on one of my slaves show its not there)

ifireball:
Ping?

jhoblitt:
Could you rebase on current master?

files/jenkins-slave.service
```diff
@@ -0,0 +1,16 @@
+[Unit]
+Description=Jenkins slave service
+Requires=network.target
+After=network.target
+
+[Service]
+Type=simple
+ExecStart=/usr/local/sbin/jenkins-slave
```
jhoblitt: Note that the path here doesn't match where the run script is being installed

jhoblitt:
I have rebased this branch on master and am working on fixing it up.  Hopefully, I'll get a PR out today.

Issue #575, opened 2016-05-05
2.1 support


cdenneen:
Does not seem to work with 2.1

``` puppet
==> jenkins2: Error: Failed to apply catalog: Execution of '/bin/java -jar /usr/lib/jenkins/jenkins-cli.jar -s http://localhost:8080 groovy /usr/lib/jenkins/puppet_helper.groovy user_info_all' returned 255: java.io.IOException: No X-Jenkins-CLI2-Port among [X-Jenkins, null, Server, X-Content-Type-Options, X-You-Are-In-Group, X-Hudson, X-Permission-Implied-By, Date, X-Jenkins-Session, X-You-Are-Authenticated-As, X-Required-Permission, Set-Cookie, Expires, Content-Length, Content-Type]
==> jenkins2:   at hudson.cli.CLI.getCliTcpPort(CLI.java:284)
==> jenkins2:   at hudson.cli.CLI.<init>(CLI.java:128)
==> jenkins2:   at hudson.cli.CLIConnectionFactory.connect(CLIConnectionFactory.java:72)
==> jenkins2:   at hudson.cli.CLI._main(CLI.java:473)
==> jenkins2:   at hudson.cli.CLI.main(CLI.java:384)
==> jenkins2:   Suppressed: java.io.IOException: Server returned HTTP response code: 403 for URL: http://localhost:8080/cli
==> jenkins2:       at sun.net.www.protocol.http.HttpURLConnection.getInputStream0(HttpURLConnection.java:1840)
==> jenkins2:       at sun.net.www.protocol.http.HttpURLConnection.getInputStream(HttpURLConnection.java:1441)
==> jenkins2:       at hudson.cli.FullDuplexHttpStream.<init>(FullDuplexHttpStream.java:78)
==> jenkins2:       at hudson.cli.CLI.connectViaHttp(CLI.java:152)
==> jenkins2:       at hudson.cli.CLI.<init>(CLI.java:132)
==> jenkins2:       ... 3 more
```

Also the "initial admin secret available in /var/lib/jenkins/secrets/..." need a way to bypass when building via puppet.

jhoblitt:
Did you get `2.1` by default from one of the managed repos?

cdenneen:
Yes from the jenkins repo `2.1-1.1`

jhoblitt:
Which platform?  It is not yet in the EL repo.

http://pkg.jenkins-ci.org/redhat-stable/

[...]

PR #576, opened 2016-05-05
Kludge around jenkins 2.x landing in the 'latest' repos

jhoblitt:
Related to #575

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,11 @@ This is intended to be a re-usable
 [Puppet](http://www.puppetlabs.com/puppet/introduction/) module that you can
 include in your own tree.
 
+# Jenkins 2
+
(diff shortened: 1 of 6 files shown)


matthewbarr:
+1

rtyler:
@jhoblitt I wonder if the repo change for an existing installation would render this as a "major" (in semver terms) incompatible change.

jhoblitt:
@rtyler Err, uh... I suppose it could cause a downgrade.  I would hope that it's not the case since presumably the latest 1.x package would obsolete the most recent lts 1.x package.

jhoblitt:
The travis failure was in the beaker tests on el6/el7 and appear to be timing (non-package) related.

PR #577, opened 2016-05-11
update puppet helper to allow changing usernames

rndmh3ro:
This PR allows the helper to check wether a user with the provided id already exists and makes it possible to update the name.

There's a test missing for this, though I'm not sure how to exactly test this. So I'd appreciate any help there.

### DIFF
--- a/files/puppet_helper.groovy
+++ b/files/puppet_helper.groovy
@@ -392,7 +392,15 @@ class Actions {
     }
 
     // Create or update the credentials in the Jenkins instance
-    def existing_credentials = util.credentials_for_username(username)
+
+    def existing_credentials
+
+    if ( id != null) {
+      existing_credentials = util.findCredentialsById(id)
+      }
+    else {
+      existing_credentials = util.credentials_for_username(username)
+    }
 
     if(existing_credentials != null) {
       credentials_store.updateCredentials(
--- a/files/puppet_helper.groovy
+++ b/files/puppet_helper.groovy
@@ -392,12 +392,11 @@ class Actions {
     }
 
     // Create or update the credentials in the Jenkins instance
-
     def existing_credentials
 
     if ( id != null) {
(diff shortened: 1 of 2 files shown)


files/puppet_helper.groovy
```diff
@@ -392,7 +392,15 @@ class Actions {
     }
 
     // Create or update the credentials in the Jenkins instance
-    def existing_credentials = util.credentials_for_username(username)
+
+    def existing_credentials
+
+    if ( id != null) {
```
jhoblitt: w/s

files/puppet_helper.groovy
```diff
@@ -392,7 +392,15 @@ class Actions {
...
+      existing_credentials = util.findCredentialsById(id)
+      }
```
jhoblitt: indention

jhoblitt:
@rndmh3ro It is difficult to "unit test" the groovy code because it doesn't function without a full up jenkins instance.  Test coverage is achieved by by the beaker acceptance tests under https://github.com/jenkinsci/puppet-jenkins/tree/master/spec/acceptance.  However, I don't believe there are existing acceptance tests that cover this method.

rndmh3ro:
I removed the w/s and correctee the indentation. However as you can see, I fail writing that test. I'm pretty new to this, so can you point me in the correct direction?
I tried adding another test [here](https://github.com/jenkinsci/puppet-jenkins/pull/577/commits/ba9ef01bb66dfe716e794fc44e12a32159d08cdf) but it does not seem to get tested at all.

rndmh3ro:
Closing this because I noticed that its supported [here](https://github.com/jenkinsci/puppet-jenkins/blob/master/NATIVE_TYPES_AND_PROVIDERS.md#jenkins_credentials)

elconas:
This one https://github.com/jenkinsci/puppet-jenkins/pull/598 fixes the same issue

PR #580, opened 2016-05-17
#574 allow newline before start of private key

br0ch0n:
This is a fix/sortof workaround for #574 

I discovered by accident that if you pad your SSH private key with a newline, the CLI helper accepts it as an argument.  Problem is that it gets added as a file path rather than direct input since startsWith doesn't handle the leading newline (but at least it gets added as SOMETHING rather than throwing an arg error).  This change sorts that out. 

P.S. I know we should be using the native types instead but I was getting puppet_x load errors when I tried it  (on 3.8, not 4+) so I went back to this.

### DIFF
--- a/files/puppet_helper.groovy
+++ b/files/puppet_helper.groovy
@@ -376,7 +376,7 @@ class Actions {
       )
     } else {
       def key_source
-      if (private_key.startsWith('-----BEGIN')) {
+      if (private_key =~ /^(\s*)-----BEGIN(.*)/) {
         key_source = new BasicSSHUserPrivateKey.DirectEntryPrivateKeySource(private_key)
       } else {
         key_source = new BasicSSHUserPrivateKey.FileOnMasterPrivateKeySource(private_key)


jhoblitt:
The acceptance tests failed -- this might be because of jenkins 2.x.  I've restarted travis.  If it fails again, could you rebase on the current master?

br0ch0n:
@jhoblitt yaa, I think master hasn't been building for awhile now: https://travis-ci.org/jenkinsci/puppet-jenkins/builds and is still red today, so I don't know that rebasing would help yet?

jhoblitt:
There have been stability issues with the docker based acceptance tests that come and go with the phases of them moon.  I don't really understand the issue -- I am getting close to marking them as allowed to fail.

br0ch0n:
went and rebased just for fun but tests have failed as expected.  Unless I missed it, I don't see any mention of using the LTS repo in the spec so I think you are testing against Jenkins 2.x as you mentioned earlier(which isn't supposed to work for whatever reason)?

br0ch0n:
any update on this one?  If it helps, I've now been using it in prod for a few months now :D

jhoblitt:
@br0ch0n Would you be willing to rebase on the current master to see if travis will pass?

GuitarrasDeAmor:
Any updates on this?

PR #583, opened 2016-05-19
Dynamic job configuration

zonArt:
Give users the possibility to perform dynamic job configuration based on concat fragment and use them to create jobs within jenkins-cli

### DIFF
--- a/manifests/job/present.pp
+++ b/manifests/job/present.pp
@@ -7,7 +7,10 @@
 # Parameters:
 #
 #   config
-#     the content of the jenkins job config file (required)
+#     the content of the jenkins job config file
+#
+#   config_file
+#     the jenkins job config file (file on disk)
 #
 #   jobname = $title
 #     the name of the jenkins job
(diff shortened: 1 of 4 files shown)


manifests/job/present.pp
```diff
@@ -7,7 +7,10 @@
 # Parameters:
 #
 #   config
-#     the content of the jenkins job config file (required)
+#     the content of the jenkins job config file (template based)
```
jhoblitt: The parameter is just a string.  No templating is applied.

manifests/job/present.pp
```diff
@@ -7,7 +7,10 @@
...
+#
+#   temp_config_path
```
jhoblitt: How about removing the 'temp_' prefix from the parameter name?  I think the name will confuse end users.

manifests/job/present.pp
```diff
@@ -29,12 +33,43 @@
   include jenkins::cli
   include jenkins::cli::reload
 
+  if $temp_config_path != undef and $config != undef {
```
jhoblitt: I believe  this could be simplified to `if $temp_config_path and $config {`.

manifests/job/present.pp
```diff
@@ -29,12 +33,43 @@
...
+  }
+
   if $jenkins::service_ensure == 'stopped' or $jenkins::service_ensure == false {
     fail('Management of Jenkins jobs requires \$jenkins::service_ensure to be set to \'running\'')
   }
 
   $jenkins_cli        = $jenkins::cli::cmd
-  $tmp_config_path    = "/tmp/${jobname}-config.xml"
+  if $config == undef {
+    $tmp_config_path = $temp_config_path
```
jhoblitt: I feel that use two variable names that are so similar is asking for trouble down the road...

manifests/job/present.pp
```diff
@@ -29,12 +33,43 @@
   include jenkins::cli
   include jenkins::cli::reload
 
+  if $temp_config_path != undef and $config != undef {
+    fail('You cannot set both $temp_config_path AND $config param, only one is required')
+  }
```
jhoblitt: A test is also needed to  make sure one of config_path or config is defined.

zonArt:
@jhoblitt Thanks for the comments, I'll take them into consideration and push an updated proposal, will try the testing in a third commit

zonArt:
Any news on this one ? Is it possible to get it merge ?

jhoblitt:
There still isn't any test coverage of the new `config_file` parameter.

zonArt:
Ok, so if I understand that well, it means I need to write a new spec/defines/jenkins_job_present.rb with all the tests related to this define ? Is it the strategy to adopt (I'm sorry for the maybe stupid question but I'm totally newb at writing puppet tests)

zonArt:
@jhoblitt, I made a test on config_file as requested, don't know if this is enough or not

edomaur:
:+1:  seems fine now :-)

edomaur:
:+1: seems fine now :-)

zonArt:
What about this PR? I guess the "needs-tests" label could be removed and what about merging it?

zonArt:
Beaker tests are now fixed thanks to @jhoblitt, could we now consider this as a merge candidate ?

zonArt:
@jhoblitt Is it possible to get this merged before new conflict occurs ? If anything is missing or wrongly done, thanks to tell so that we can go further on with it

edomaur:
+1

PR #584, opened 2016-05-19
Removing 'enable' statement on job.pp and job/present.pp

zonArt:
Removing 'enable' statement on job.pp and job/present.pp which are breaking idempotence

### DIFF
--- a/manifests/job.pp
+++ b/manifests/job.pp
@@ -18,6 +18,9 @@
 #   jobname = $title
 #     the name of the jenkins job
 #
+#   enabled
+#     deprecated parameter (will have no effect if set)
+#
 #   ensure = 'present'
 #     choose 'absent' to ensure the job is removed
 #
@@ -29,6 +32,7 @@ define jenkins::job(
   $source   = undef,
   $template = undef,
   $jobname  = $title,
+  $enabled  = undef,
   $ensure   = 'present',
   $difftool = '/usr/bin/diff -b -q',
 ){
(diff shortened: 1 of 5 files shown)


jhoblitt:
@zonart I confess to being a bit torn because I would like to see this feature removed.  However, I think we can't remove the parameter without a major API bump. 

@jenkinsci/puppet-jenkins-developers how would you feel about making this param a no-op and adding a warning?

zonArt:
@jhoblitt I made it a no-op with warning as you suggested, I agree that this is a better compromise

jhoblitt:
@jenkinsci/puppet-jenkins-developers Could one of the other maintainers ack this PR?  I am in favor of this change because the feature, as implemented, is inherently non-indempotent and fragile as there is no semantic parsing of the xstream dump.

zonArt:
Any news on this one ? Is it possible to get it merge ?

edomaur:
:+1:  if we can merge it soon it would be really nice

cClaude:
continuous-integration/travis-ci/pr — The Travis CI build failed

Seem to be link to a docker issue, not related to this PR.

cClaude:
+1

zonArt:
Tests are fixed, parameter has been set as deprecated, backward compatibility is not broken as feared at first commits, what else is missing? If there are any concerns remaining, please tell so that I can propose something

pouete:
+1

elconas:
Do you do semantic versioning for this puppet module ? If so, I think this breaks users as it changes behaviour and needs a major version change. 

In our case we rely on this feature to create disabled jobs and let the build manager decide when and when to run it as this job, bootstraps a whole CI Pipeline with JobDSL (we added seed_only option as shown in https://github.com/jenkinsci/puppet-jenkins/pull/500).

So with the current implementation, behaviour changes silently (ok a warning, but no failure). I would rather see a failure to force the admin to look into it and break no assumptions.

PR #587, opened 2016-05-20
Improve idempotency of credentials define

br0ch0n:
the CLI is flaky, so the 'unless' needs retries as well otherwise we end up unnecessarily reinstalling creds (and reloading jenkins) the first time the CLI doesn't respond due to things like plugin installs and previous restarts/reloads.

### DIFF
--- a/manifests/credentials.pp
+++ b/manifests/credentials.pp
@@ -22,12 +22,16 @@ define jenkins::credentials (
   $private_key_or_path = '',
   $ensure = 'present',
   $uuid = '',
+  $unless_tries = 20,
+  $unless_try_sleep = 5,
 ){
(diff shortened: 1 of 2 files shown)


manifests/credentials.pp
```diff
@@ -22,12 +22,16 @@
   $private_key_or_path = '',
   $ensure = 'present',
   $uuid = '',
+  $unless_tries = 20,
+  $unless_try_sleep = 5,
```
jhoblitt: Instead of adding new params, how about reusing `$::jenkins::cli_tries` & `$::jenkins::cli_try_sleep` as is done in `jenkins::cli::exec`?

Issue #572, opened 2016-04-29
Update jenkins::slave::labels to be an array.


madelaney:
I'd like to suggest that the jenkins::slave::labels param be updated to either take an array or a string. 

The reason for this is our setup uses hiera and we've found that using a string is kind of messy. Using an array and being able to use deep_merge is extremely helpful.

I'm happy to create a pull request if you're open to this.

jhoblitt:
@madelaney +1  I completely agree.

jhoblitt:
To be pedantic, I would be more than happy to accept a PR to implement this proposal.

jhoblitt:
Resolved by #591

PR #590, opened 2016-05-24
-- #572: Converted the jenkins slave labels param to an array

madelaney:


jhoblitt:
My interpretation of the proposal in #572 was that the `labels` param should accept either a string or an array, not just an array, and that a string would continue to be handled in a backwards compatible manner.

As it currently stands, this PR would require a major version bump.

madelaney:
@jhoblitt  You are correct, my apologies. I'll make the necessary changes.

PR #592, opened 2016-05-27
enables the use of a template for the plugin configuration

witjoh:
Installing plugins using hiera and the jenkins::plugins class, it is impossible to generate the config_content using a template.  With this addition, it becomes a piece of cake.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ README.pdf
 .ruby-*
 *.sw*
 spec/fixtures
+!spec/fixtures/modules/site_jenkins
 Gemfile.lock
 .vagrant*
 .bundle
--- a/manifests/plugin.pp
+++ b/manifests/plugin.pp
@@ -5,10 +5,17 @@
 # config_content = undef
 #   Content of the config file for this plugin. See also $use_template
 #   when the use of a template is desired.
+#   Supported formats :
+#   * $config_content => '<module>/plugin_config.xml.erb'
+#   * $config_content => '<module>/plugin_config.xml.part1.erb, '<module>/plugin_config.xml.part2.erb'
+#   * $config_content => '<module>/plugin_config.xml.epp'
+#   * $config_content => ['<module>/plugin_config.xml.epp', { 'param1' => 'value2', 'param2' => 'value2' } ]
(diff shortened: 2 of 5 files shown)


jhoblitt:
What is the use case for this feature that isn't covered by `config_content  => template(...)`?  Also, puppet 4.x users may want to use `epp()`.

witjoh:
about the epp, thats indeed a good point, will add support for it. (just need to read a bit, since we are still on 3.x).

when one puts jenkins::plugin_hash in hiera, it is impossible to get the content from a template into the config_content.  We could indeed call the plugin resource for those plugins that need a template configuration, but then we have to keep the configuration in two places, in hiera and in our jenkins profile. 
Keeping everything in one place is what we prefer.
See also this ticket :
https://tickets.puppetlabs.com/browse/HI-295

Will update this PR with epp support today.

witjoh:
* added support for both erb and epp
* added also rspec testing

witjoh:
@jhoblitt Any change to have a quick review ?

sdiepend:
@jhoblitt @rtyler @witjoh Any updates on this? I'm also interested in this pull request.

jhoblitt:
Haven't had any mental bandwidth.  I'm surprised that the `epp()` function call isn't causing a problem with puppet 3.8 -- anyone have an explanation?

witjoh:
@jhoblitt 
Did a little test with 3.8.5, and I get the following error : 
Error: epp() is only available when parser/evaluator future is in effect at /home/vagrant/modules/epptest/manifests/init.pp:7 on node slave.vagrant.com

[vagrant@slave ~]$ puppet agent --configprint parser
current
[vagrant@slave ~]$ puppet apply --modulepath modules modules/epptest/examples/init.pp  --parser future does work as expected.

witjoh:
@jhoblitt ping ?

witjoh:
@jhoblitt :  is this still in the running ? time for a rebase ?

jhoblitt:
@witjoh Could you try rebasing this PR to see if the test will pass?  The acceptance tests are now reliable on the master branch.

bastelfreak:
Hello. I'm going to close this PR due to inactivity. Please reopen and rebase it if you're still interested in it.

PR #595, opened 2016-06-08
Avoid Plugin installation loops 

elconas:
Commit 9bf2a1fa7eb66b06b6cb9c8f659019a0dcb46c30 introduced cleanup logic for installed modules. However when you install plugins from jenkins-ci.org (without source => '' parameter), auto detection of jpi/hpi extension does not work (it is always assumed hpi) and this leads to plugin installation loops. 

begin:
  Puppet removes existing .jpi files
  Puppet installs Plugin plugin.hpi file
  Puppet restarts Jenkins
  Jenkins renames .hpi to .jpi 
goto begin:

This fix allows specifying the plugin extension type manually, avoiding those loops.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,6 +131,21 @@ can specify a checksum. You can also define a checksum type with
   }
 ```
 
+#### JPI Plugins
+
+When installing jpi plugins, you must make sure that puppet knows the
+extension of the module, otherwise this jenkins module will cleanup 
+and install the plugin on each puppet run. 
+
+```puppet
+  jenkins::plugin { 'cvs':
+    version       => '2.12',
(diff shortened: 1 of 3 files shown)


jhoblitt:
Could you provide an example of a plugin that causes the failure mode your describing?  I haven't seen this problem myself.

elconas:
Try to install the plugins: 

  matrix-project (JPI)
  nodelabelparameter (HPI)

jhoblitt:
I'm going to close this since I don't think this is a concern with the public updatecenter.  The complete URL could always be passed via the `source` param as well.

PR #596, opened 2016-06-08
Add support to specify multiple jenkins credentials with the same username

elconas:
Sometimes it is needed to have multiple credential sets in Jenkins using the same username but different passwords. This is the case, when using JFrog artifactory, here you need encrypted passwords to access artifactory, while you need plain plassword to access stash. (Both services AD integrated as common in larger enterprises). 

This fix adds support to specify multiple credentials with same username but different uuid. It also fixed the remove logic (making it idempotent) and fixes the matching for usernames (more specifiy grep expression).

Same result that is fixed by this commit can be achieved using the new experimental resource type jenkins_credentials, however this  resource type does not work on PuppetServer using jruby as due to puppet_x library location (https://tickets.puppetlabs.com/browse/SERVER-973).

### DIFF
--- a/files/puppet_helper.groovy
+++ b/files/puppet_helper.groovy
@@ -74,6 +74,17 @@ class Util {
     return CredentialsMatchers.firstOrNull(allCreds, idMatcher)
   }
 
+  def findCredentialsByNameOrId(String username, String id="") {
+    def existing_credentials
+ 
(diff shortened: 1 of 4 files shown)


elconas:
Abort and reopen with rebase in https://github.com/jenkinsci/puppet-jenkins/pull/598

PR #599, opened 2016-06-08
Escape +'s when grepping through jenkins plugin version numbers

cliff-svt:
For some reason, the maintainers of the jenkins "Build Monitoring Plugin" (https://wiki.jenkins-ci.org/display/JENKINS/Build+Monitor+Plugin) decided to put a "+" in their version names (e.g., "1.9+build.201606052339"). This confuses the logic that uses grep to search for installed plugins. The "+" is interpreted as a special regexp character, so no match is found, and puppet tries to install the plugin (and restart jenkins) over and over again. This fix escapes +'s in the search string before it's passed to grep.

### DIFF
--- a/manifests/plugin.pp
+++ b/manifests/plugin.pp
@@ -70,7 +70,14 @@ define jenkins::plugin(
       default => $update_url,
     }
     $base_url = "${plugins_host}/download/plugins/${name}/${version}/"
-    $search   = "^${name} ${version}$"
+    # Escape +'s in $version when constructing $search.
+    # * We can't use single quotes for the replacement string because
+    #   puppet 3 and puppet 4 interpret '\\' differently.
+    # * We can't use double quotes without a variable interpolation or
+    #   lint complains.
+    $empty    = ''
+    $escver   = regsubst ($version, '\+', "${empty}\\\\+", 'G')
+    $search   = "^${name} ${escver}$"
   }
   else {
     $plugins_host = $update_url ? {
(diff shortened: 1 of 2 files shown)


jhoblitt:
Oh, that is lovely.  Not sure why this change is failing under 4.x.

cliff-svt:
Double backslash is a single quote escape string in puppet 4 but not puppet 3:
* https://docs.puppet.com/puppet/latest/reference/lang_updating_manifests.html#check-single-quoted-strings-for-double-backslashes

cliff-svt:
Now it's failing the lint test ("double quoted string containing no variables").  Sigh.

cliff-svt:
Latest failure:
  Error: Failed to apply catalog: Execution of '/usr/bin/java -jar /usr/lib/jenkins/jenkins-cli.jar -s http://localhost:8080 groovy /usr/lib/jenkins/puppet_helper.groovy job_list_json' returned 255: java.io.IOException: No X-Jenkins-CLI2-Port among [X-Jenkins, null, X-Hudson, X-Hudson-Theme, Content-Length, Expires, X-Jenkins-Session, Set-Cookie, Content-Type, Server, Cache-Control, X-Content-Type-Options]

Looks like an infrastructure problem maybe?

jhoblitt:
Errors in the acceptance tests can be ignored.  There have been odd issues with running docker containers under travis and I think fact changes may be broken things.  Haven't had any free time to try to debug it.

jhoblitt:
I'd say it looks good but that is certainly ugly. ;)

PR #567, opened 2016-04-22
Add param disableClientsUniqueId

xdexter:
Adds the ability to set the parameter "disableClientsUniqueId".

PR #600, opened 2016-06-08
Support multiple swarm client args 

alvin-huang:
My desire was to use the -disableClientUniqueId option similar to #567. I thought it may be better to just create a long string for users to input multiple swarm client arguments. The list of options are listed here: https://github.com/jenkinsci/swarm-plugin/blob/master/client/src/main/java/hudson/plugins/swarm/Options.java.

There are quite a number of options that aren't implemented into the module currently and ultimately, I think this would be an easier way to do it than to add class parameters for each option. 

Thoughts?


jhoblitt:
Could you add some rspec coverage of this change and rebase on current master?

alvin-huang:
@jhoblitt Rebased and rspec tests added.

jhoblitt:
Resolved by #665

alvin-huang:
@jhoblitt you mean #655 :)

Issue #573, opened 2016-04-29
Update jenkins::slave::java_args to be an array.


madelaney:
I'd like to suggest that the jenkins::slave::java_args param be updated to either take an array or a string. 

The reason for this is our setup uses hiera and we've found that using a string is kind of messy. This is mostly because we have a distributed environment and we want to adjust some arguments based on locale but we want to keep the others uniform.

I'm happy to create a pull request if you're open to this.

jhoblitt:
:+1:

PR #604, opened 2016-06-20
-- #573: Convert jenkins::slave::java_args to support both strings and arrays.

madelaney:
This change will allow users to use 'jenkins::slave::java_args' as a string, or array. With support for hiera_array merging; based on hiera config.


jhoblitt:
Looks good.  Thank you for contributing!

PR #606, opened 2016-06-30
Adding LDAP as a user realm.

sdiepend:
By setting the user_realm to ldap and passing in the necessary config fields in ldapconfig through hiera for example then puppet will manage the ldap config.

Parameters to be set in the $ldapconfig hash:

    server                     => "",
    rootDN                     => "",
    userSearch                 => "",
    managerDN                  => "",
    managerPasswordSecret      => "",
    inhibitInferRootDN         => "",
    disableMailAddressResolver => "",
    displayNameAttributeName   => "",
    mailAddressAttributeName   => ""

These will be loaded into a json file and then used by the puppet_helper groovy script to pass these arguments to the LDAPSecurityRealm constructor function.

### DIFF
--- a/files/puppet_helper.groovy
+++ b/files/puppet_helper.groovy
@@ -21,9 +21,12 @@ import com.cloudbees.plugins.credentials.common.*
 import com.cloudbees.plugins.credentials.domains.*
 import com.cloudbees.plugins.credentials.impl.*
 import groovy.transform.InheritConstructors
+import groovy.json.JsonSlurper
 import hudson.model.*
 import hudson.plugins.sshslaves.*
 import hudson.util.*
+import hudson.security.LDAPSecurityRealm
+import hudson.security.HudsonPrivateSecurityRealm
 import jenkins.model.*
 import jenkins.security.*
 import org.apache.commons.io.IOUtils
@@ -36,6 +39,7 @@ class UnsupportedCredentialsClass extends Exception {}
 class InvalidCredentialsId extends Exception {}
 @InheritConstructors
 class MissingRequiredPlugin extends Exception {}
+class CanNotLoadJsonMapFileException extends Exception{}
 
 ///////////////////////////////////////////////////////////////////////////////
 // Util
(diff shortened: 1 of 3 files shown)


alberto-sierra:
I recently worked on a similar groovy script and was not able to find a way to provide an encrypted password in the groovy script or properties ($ldapconfig hash) before hand. Is the `managerPasswordSecret` in the hash in plain text?
Also, just a heads up that `LDAPSecurityRealm` is part of the `ldap` plugins that comes installed by default on 1.x versions but it doesn't on 2.x and will probably break if the plugin is not installed before running the groovy script.

sdiepend:
@alberto-sierra 
1) We use hieradata and you could use eyaml for encrypted hiera data. But I'm not sure if you would still be able to see it in the helper groovy script, haven't looked at that yet.

2) Ok, I'll have to put in a requirement then and make it 2.x proof.

3) An extra remark: Once the LDAP has been configured, you'll have to find a way to talk to the jenkins-cli without the "Anonymous user". You can add a public key to a existing user with the rights to talk to the cli. And the private key should be on the jenkins master where puppet will run and execute the groovy scripts via the cli. The funny thing is you can add the pub key to any user you want it does not do an identity check and will just check with all existing pub keys.

alberto-sierra:
@sdiepend I'm using `FullControlOnceLoggedInAuthorizationStrategy` with LDAP and I tried to add a pubkey to an existing user like you mentioned (3) but the CLI still doesn't work (`java.io.EOFException`) at least in v2.x. I will try this again in a 1.x install and see if something changed or i'm doing something wrong.

sdiepend:
@alberto-sierra I will test it on 2.x the beginning of next week, I'll let you know how it goes.

alberto-sierra:
@sdiepend I was able to make the `cli` work with your "hack". Since a valid service account is required for LDAP binding, I used this same account for the `jenkins_cli` key. While I didn't use the groovy_helper script from this module, I believe you could use the `create_or_update_user` function to add the public key after LDAP has been configured.

sdiepend:
@alberto-sierra that's great news, I'm sorry, haven't been able to dedicate much time to this. But won't create_or_update_user need a key already after LDAP is configured?

PR #611, opened 2016-07-07
jenkins 2.x support

jhoblitt:


jhoblitt:
The acceptance tests for el6 & el7 are passing for me locally but ubuntu fails due to the package installation starting up the daemon with the wizard enabled.  I think we may be stuck unless the debian packaging is changed.  I'm baffled as to why ubuntu is passing under travis.

jhoblitt:
Ubuntu is now failing as expected... very odd that it was somehow passing briefly.  In any event, the only method I can think of to support the debian packages is to detect when the wizard has been run on an install and attempt to use the auto-generated passphrase.  Does anyone have a better idea?

jhoblitt:
@jenkinsci/puppet-jenkins-developers What do you folks think about merging this?  It seems to generally work for bootstrapping redhat but fails for ubuntu due to the debian packages auto-starting the daemon (oddly, the acceptance tests sometimes pass under travis... all I can think is that some times a jenins 1.x package is installed).

I have split out several attempts to make the beaker tests reliable under travis and merged them separately (#607, #613 , & #615).  I am at a loss to explain these problems -- I suspect there is some sort of random/screwy network problem when running docker containers on travis' infrastructure.

PR #612, opened 2016-07-11
add 'proxy_server' param to jenkins::slave class

jhoblitt:
The `jenkins` and `jenkins::slave` classes should be fully decoupled so
that the `jenkins` class is not required to be in the catalog for a
"slave only" node.


matthewbarr:
Makes sense..

PR #622, opened 2016-08-16
Pin the loose dependency (from hiera) on json_pure to something less than 2.0.2

rtyler:
2.0.1 requires Ruby version: >= 0

2.0.2 requires Ruby version: ~> 2.0

FUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUUU

https://twitter.com/agentdero/status/765587955612160001

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,7 +12,7 @@ gem 'metadata-json-lint'
 gem 'retries', '~> 0.0.5'
 gem 'travis', '~> 1.8'
 gem 'parallel_tests'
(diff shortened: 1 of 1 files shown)


rtyler:
@jhoblitt I think this PR is good. The beaker tests will not setup right, but I think that's unrelated to this change

jhoblitt:
agreed.  The beaker tests have been completely unreliable on top of travis.

PR #626, opened 2016-08-19
Plugins fail to install due to search false positives

m3brown:
This change brings the regex pattern for plugin (without version) search inline with the regex pattern for plugin+version search on line 80.

Without the prepended caret (^), plugins that are not installed may fail to install due to false positives.  Example:

* Install the `ssh-credentials` plugin
* Install the `credentials` plugin

With the current logic, the regex "credentials" will match the existing plugin "ssh-credentials" and the plugin will not be installed.  Adding a ^ fixes this issue.

### DIFF
--- a/manifests/plugin.pp
+++ b/manifests/plugin.pp
@@ -85,7 +85,7 @@ define jenkins::plugin(
       default => $update_url,
     }
     $base_url = "${plugins_host}/latest/"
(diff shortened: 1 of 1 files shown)


jhoblitt:
I think the failure was random travis glitching.  I've kicked it.

PR #632, opened 2016-09-01
-- 616: adding support for passing arbitrary arguments to the java swarm c…

madelaney:
This allows the end-user to give an arbitrary list of additional arguments to the swarm client.

### DIFF
--- a/manifests/slave.pp
+++ b/manifests/slave.pp
@@ -79,6 +79,11 @@
 #   parameter so the `::jenkins` class does not need to be the catalog for
 #   slave only nodes.
 #
+# [*slave_args*]
+#
+#   This takes a string, or an array, of additional arguments to pass into the Jenkins
+#   swarm client. Please see your version of the Swarm client for available options.
+#
 
 # === Examples
 #
@@ -119,6 +124,7 @@ class jenkins::slave (
   $source                   = undef,
   $java_args                = undef,
   $proxy_server             = undef,
+  $slave_args               = undef,
 ) inherits jenkins::params {
   validate_string($slave_name)
   validate_string($description)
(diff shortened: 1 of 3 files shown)


manifests/slave.pp
```diff
@@ -170,6 +176,16 @@
     }
   }
 
+  if $slave_args {
+    if is_array($slave_args) {
+      $_combined_slave_args = hiera_array('jenkins::slave::slave_args', $slave_args)
```
jhoblitt: Whoa -- I feel this is pretty surprising behavior.  What is the motivation for doing this?

madelaney: @jhoblitt fair question. The idea behind this is that there are flags, like `disableClientsUniqueId` and `deleteExistingClients`, that I want to send to the client. I thought of listing these out explicitly but felt this was a bit easier to manage going forward.

rtyler: I don't understand the need for a Hiera lookup though, you can already inject class parameters by defining:

```yaml
jenkins::slave::slave_args: "bippidyboppidy-boop"
```

madelaney: @rtyler , I did that so we can take advantage of the deep merge behavior.

rtyler: @madelaney I figured that might be the case, but I think that makes this behavior even more confusing. Hiera already has configuration (which I *think* applies to injected class parameters) for its [merge behaviors](https://docs.puppet.com/hiera/3.2/configuring.html#mergebehavior) (also [array merge handling](https://docs.puppet.com/hiera/3.2/lookup_types.html#array-merge)).

There's also [this documentation](https://docs.puppet.com/puppet/latest/reference/lookup_quick.html#specifying-merge-behavior) which might be relevant? The myriad of merge-related documents annoys and confuses me :person_frowning: 


Anyways, I generally find accepting class parameters *plus* doing a `hiera_array` lookup of the same key to be weird.

madelaney: @rtyler, fair enough. The docs can be very confusing. If I change line 181-182 to read more like:
```
$_real_slave_args = join($slave_args, ' ')
```
Would you be okay with that?

PR #646, opened 2016-09-30
Move jenkins::params::default_plugins to class param

jeffmccune:
Without this patch the configuration does not converge.  The credentials plugin
is downloaded and installed every configuration run.  This patch addresses the
problem by making it possible to override the default configuration of the
credentials plugin with the following hiera data:

```
jenkins::params::default_plugins: []
jenkins::plugin_hash:
  credentials:
    version: 2.1.5
    digest_string: 7db002e7b053f863e2ce96fb58abb98a9c01b09c
    digest_type: sha1
```

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -1,7 +1,9 @@
 # Class: jenkins::params
 #
 #
-class jenkins::params {
+class jenkins::params(
+  $default_plugins = [ 'credentials' ], # required by puppet_helper.groovy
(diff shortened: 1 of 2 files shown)


jhoblitt:
It looks like I goofed up and never 'posted' my review comments.  I'm not enthusiastic about adding parameters to the params class. #675 is more to my taste but needs some tests.

PR #648, opened 2016-10-06
Avoid "Undefined variable 'proxy_server'" warnings.

jgreen210:
These changes avoid one or many warnings:

Warning: Undefined variable 'proxy_server'; 
 (file & line not available)

...from direct_download.pp and plugin.pp. I was seeing warnings from direct_download.pp when using direct_download but not using a proxy. I'm using puppet 4.7.0.


jgreen210:
The test failures don't appear to be caused by my changes.

PR #650, opened 2016-10-12
Disable swarm unique id on slave

mterzo:
CI broke on older ruby, net-http-persistent 3.0 requires ruby 2.1, adding this change as well.


jhoblitt:
I've restarted the failed travis job.  It looks like the build slave went away mid-output.

templates/jenkins-slave-defaults.erb
```diff
@@ -60,6 +60,7 @@ JENKINS_USERNAME=<%= @quoted_ui_user %>
 JENKINS_PASSWORD=<%= @quoted_ui_pass %>
 
 OTHER_ARGS="<%= '-disableSslVerification' if @disable_ssl_verification -%>"
+DISABLE_UNIQUE="<%= '-disableClientsUniqueId' if @disable_ssl_verification-%>"
```
rtyler: I believe this is using the wrong variable in the ERB template. Instead of `disable_ssl_verification` it should be `disable_clients_unique_id`

Issue #652, opened 2016-10-17
Labels under macOS are over escaped.


madelaney:
When using the jenkins-slave on macOS, if I supply more then a single label the variable `LABELS` is over escaped/quotes by the time it gets to setting `JENKINS_SLAVE_ARGS`

For example, if I set `LABELS="hello world"` with the current version, I get the following `JENKINS_SLAVE_ARGS`:
``` 
JENKINS_SLAVE_ARGS='-mode normal -executors 1 -username jenkins -password ***** -name 10-70-7-230 -master https://jenkins -labels '\''hello world'\'' -fsroot /Users/build/.jenkins    -deleteExistingClients -disableClientsUniqueId '
```

If you notice the lables are expanded to `'\''hello world'\''` which causes the jar command to look for a argument named `hello` and `world`.

madelaney:
After some usage at my work, I found my initial version kind of sucked; it didn't always work.

I've been playing around with this version but wanted to get a barometer of what people thought.
```
<% if @_real_labels.is_a? String %>
LABELS="<%= @_real_labels.split.map.each { |item| "-labels #{item}" }.join(' ') unless @_real_labels.nil? %>"
<% else %>
LABELS="<%= @_real_labels.map.each { |item| "-labels #{item}" }.join(' ') unless @_real_labels.nil? %>"
<% end %>
```

This will end up generating a string of labels that are all prefixed with '-labels' to use the fact that the swarm client accepts multiple inputs of '-labels'.

jhoblitt:
@madelaney I suspect that this issue was resolved as a side effect of #721. Could you test from current master?

madelaney:
@jhoblitt I tried to test this but I ran into some auth errors. I'll try to get an update as soon as I can.

jhoblitt:
@madelaney Doh! `start-slave.sh.erb` probably needs to be updated to handle the new sysconf file format...

madelaney:
Maybe. I might be a bit unique in that I take advantage of the syntax `@<file_name>` syntax.

I noticed that when I tried to upgrade, the [password param for the slave](https://github.com/jenkinsci/puppet-jenkins/blob/master/templates/jenkins-slave-defaults.erb#L50) has the correct string (i.e. @/etc/secure/somefilename) but the service that started Jenkins always tried to export/set [passwordEnvVariable](https://github.com/jenkinsci/puppet-jenkins/blob/master/templates/jenkins-slave-run.erb#L31) which is not yet set.

PR #653, opened 2016-10-17
#652: Adjust how the LABELS_ARG variable gets set to avoid over quoting.

madelaney:
When setting the variable `LABELS_ARG` strip quotes from the the variable `LABELS` to avoid over quoting.


madelaney:
Found an issue with this edit, going to close to re-work.

PR #661, opened 2016-11-01
Add support for disabling management of Service[jenkins]

rtyler:
This is especially useful when deploying Jenkins from a Docker container, but
still wishing to use some of the great abstractions around job/user/etc
management in this module.

This doesn't change the expectation that /something/ will implement
Service[jenkins] in the catalog, as that would make the implementation of types
like jenkins::cli::exec obscenely complex (and maybe not even possible).


jhoblitt:
@rtyler Sorry for the long delay!

PR #663, opened 2016-11-01
adding service_provider for EL7 for jenkins-slave service

alvin-huang:
For consistency sake, if we are using service_provider for the `jenkins` master service, we should probably add the ability to use it for the `jenkins-slave` service on the slaves as well.

### DIFF
--- a/manifests/slave.pp
+++ b/manifests/slave.pp
@@ -90,6 +90,9 @@
 #   parameter so the `::jenkins` class does not need to be the catalog for
 #   slave only nodes.
 #
+# [*service_provider*]
+#   Service provider for jenkins-slave service.
+#
 
 # === Examples
 #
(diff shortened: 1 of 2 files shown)


jhoblitt:
Does this resolve an issue you are experiencing?  The service_provider nonsense came about due to a core puppet problem that should be resolved for puppet > 4.5.

alvin-huang:
IIRC--On CentOS7 boxes, the module drops a systemv init script for the jenkins-slave service and you either need to reload systemd or tell Puppet to treat it like it would for CentOS6 (by the provider).

alvin-huang:
Upon review of #692 that probably solves this issue...

jhoblitt:
@alvin-huang That is good news.  #692 isn't quite ready yet but I hope to merge it today.

Issue #668, opened 2016-11-05
'digest_type' default of 'sha1' is causing all plugins to install repeatedly


nick-george:
This could be an issue with the puppet/archive module rather than this module.

I'm using the latest versions of both this and the archive modules and each time I run puppet,  all jenkins plugins are re-downloaded and re-installed. 

    Notice: /Stage[main]/Jenkins::Config/Jenkins::Plugin[credentials]/Archive[credentials.hpi]/ensure: replace archive: /var/lib/jenkins/plugins/credentials.hpi from (sha1)91bd47ae96be9ebe199beb820296d2633994dfb0 to (sha1)
    Info: /Stage[main]/Jenkins::Config/Jenkins::Plugin[credentials]/Archive[credentials.hpi]: Scheduling refresh of Service[jenkins]
    Notice: /Stage[main]/Jenkins::Config/Jenkins::Plugin[credentials]/File[/var/lib/jenkins/plugins/credentials.hpi]/owner: owner changed 'root' to 'jenkins'
    Notice: /Stage[main]/Jenkins::Config/Jenkins::Plugin[credentials]/File[/var/lib/jenkins/plugins/credentials.hpi]/group: group changed 'root' to 'jenkins'

I have managed to work around it by manually setting "digest_type" to "none" for all plugins, but this doesn't work for the credentials plugin, which is being installed out of my control. 

The fix mentioned in #665 would allow me to work around the problem I'm experiencing.

elconas:
Fixed by https://github.com/jenkinsci/puppet-jenkins/pull/672

PR #672, opened 2016-11-18
Fix https://github.com/jenkinsci/puppet-jenkins/issues/668

elconas:
This is a Fix/Workaround for voxpupuli/puppet-archive bug https://github.com/voxpupuli/puppet-archive/issues/242

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -16,6 +16,7 @@ gem 'rubocop', '< 0.42.0'
 
 gem 'json_pure', '< 2.0.2'
 gem 'net-http-persistent', '<= 2.9.4', :require => false if RUBY_VERSION < '2.0.0'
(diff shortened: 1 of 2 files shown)


elconas:
gem addressable (used by travis indirectly) has release 2.5.0 which now needs some other gem which requires ruby 2.0. Second commit fixes this

Gemfile
```diff
@@ -16,6 +16,7 @@ gem 'rubocop', '< 0.42.0'
 
 gem 'json_pure', '< 2.0.2'
 gem 'net-http-persistent', '<= 2.9.4', :require => false if RUBY_VERSION < '2.0.0'
+gem 'addressable', '<= 2.4.0', :require => false if RUBY_VERSION < '2.0.0'
```
jhoblitt: Please remove this as ruby < 2 has been dropped from the travis matrix.

Issue #673, opened 2016-11-18
setting security is not idempotent


elconas:
Currently on every run the security is configued again and again, causing a reload of jenkins.

roidelapluie:
:-1: This is not working

PR #674, opened 2016-11-18
Fix security setting idempotentance

elconas:
fixes https://github.com/jenkinsci/puppet-jenkins/issues/673

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -16,6 +16,7 @@ gem 'rubocop', '< 0.42.0'
 
 gem 'json_pure', '< 2.0.2'
 gem 'net-http-persistent', '<= 2.9.4', :require => false if RUBY_VERSION < '2.0.0'
+gem 'addressable', '<= 2.4.0', :require => false if RUBY_VERSION < '2.0.0'
 
 group :development do
   gem 'simplecov'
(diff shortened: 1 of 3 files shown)


roidelapluie:
:-1: This is not working; still reload is triggirid

Gemfile
```diff
@@ -16,6 +16,7 @@ gem 'rubocop', '< 0.42.0'
 
 gem 'json_pure', '< 2.0.2'
 gem 'net-http-persistent', '<= 2.9.4', :require => false if RUBY_VERSION < '2.0.0'
+gem 'addressable', '<= 2.4.0', :require => false if RUBY_VERSION < '2.0.0'
```
elconas: To fix spec errors as the latest adressable (2.5.0) needs ruby 2.0 and then the 1.9 integration tests on Travis failes.

rtyler:
@jhoblitt this looks reasonable to me, any downsides to merging?

Gemfile
jhoblitt: Please remove this.  Ruby < 2 has been dropped from the travis matrix.

jhoblitt:
@rtyler The Gemfile change should be dropped.  Otherwise, it looks good to me.

Issue #665, opened 2016-11-03
Credentials plugin hard coded


madAndroid:
Hi,

We set a specific version of the credentials plugin, and set that version in our profile module - I've just updated to the latest version of `rtyler/jenkins` and it looks as though this plugin is now being hard coded into the `params.pp`, which means we can no longer control this plugin via our profile module (we're getting duplication error during testing). Is there any way we can make this plugin be optional, defaulting to being installed, and then we can override it in our profile module if we need to?

Thanks in advance.

elconas:
Same issue here. We install jenkins plugins with checksum and version pinned with code around rtyler/jenkins. Now with 1.7.0 credentials plugin beeing managed by rtyler/jenkins and jenkins::params::default_plugins not beeing parametrizable, every puppet run restarts Jenkins, causing long running builds to die ! 

(Again, I don't think this puppet module does semantic versioning right, sorry for the blame)

```
Notice: /Stage[main]/Jenkins::Config/Jenkins::Plugin[credentials]/Archive[credentials.hpi]/ensure: replace archive: /opt/jenkins/data/plugins/credentials.hpi from (sha1)49f4e6386d284bf8468e15bcd279c4a64a37f9ab to (sha1)
Info: /Stage[main]/Jenkins::Config/Jenkins::Plugin[credentials]/Archive[credentials.hpi]: Scheduling refresh of Service[jenkins]
Notice: Notice: /Stage[main]/Jenkins::Config/Jenkins::Plugin[credentials]/File[/opt/jenkins/data/plugins/credentials.hpi]/owner: owner changed 'root' to 'jenkins'
Notice: /Stage[main]/Jenkins::Config/Jenkins::Plugin[credentials]/File[/opt/jenkins/data/plugins/credentials.hpi]/group: group changed 'root' to 'jenkins'
Notice: /Stage[main]/Jenkins::Config/Jenkins::Plugin[credentials]/File[/opt/jenkins/data/plugins/credentials.hpi]/mode: mode changed '0640' to '0644'
Notice: /Stage[main]/Jenkins::Service/Service[jenkins]: Triggered 'refresh' from 1 events
```

elconas:
Workaround provided by https://github.com/jenkinsci/puppet-jenkins/pull/675
(set default_plugins = [] and manage yourself)

jhoblitt:
[...]

PR #675, opened 2016-11-18
Allow Workaround for #665

elconas:
Allow users to overwrite the default_plugins and manage them "by themselves" (with pinning etc.)


Gemfile
```diff
@@ -16,6 +16,7 @@ gem 'rubocop', '< 0.42.0'
 
 gem 'json_pure', '< 2.0.2'
 gem 'net-http-persistent', '<= 2.9.4', :require => false if RUBY_VERSION < '2.0.0'
+gem 'addressable', '<= 2.4.0', :require => false if RUBY_VERSION < '2.0.0'
```
jhoblitt: ???

PR #687, opened 2016-12-21
fix exclusion of puppet 3.x / ruby 2.2

jhoblitt:


jhoblitt:
@rtyler FYI - looks like it was yaml float conversion magic breaking the matrix exclusion.

PR #689, opened 2016-12-23
add ::jenkins::sysconfdir parameter

jhoblitt:
Controls the path to the "sysconfig" file that stores jenkins service
startup variables


jhoblitt:
@rtyler This is foundational refactoring for systemd support.

jhoblitt:
@rtyler I know it is the holiday slow down but I want to get a PR or two out today that depends on this so I'm going to go ahead and merge it.  We can always back it out later...

PR #694, opened 2016-12-30
systemd jenkins master support on RedHat

jhoblitt:


manifests/systemd.pp
```diff
@@ -0,0 +1,52 @@
...
+  # transition can not set a prior_to on
+  # Systemd::Unit_file['jenkins-slave.service'] as it is not a native type so
+  # we must us the sysv init script as a proxy
+  file { $sysv_init:
+    ensure                  => 'absent',
+    # XXX if this is not set, the seluser property will claim is it out
+    # of sync and the transition resource will always fire.  It isn't
+    # clear if this is a bug or a feature of the file resource.
+    selinux_ignore_defaults => true,
+  } ->
```
rtyler: This is kind of  weird way of specifying the dependency, and reason `systemd::unit_file` doesn't have a `require` parameter?

manifests/systemd.pp
```diff
@@ -0,0 +1,52 @@
...
+  $sysv_init = "/etc/init.d/${service}"
+
+  file { "${libdir}/${service}-run":
+    content => template("${module_name}/${service}-run.erb"),
+    owner   => $user,
+    mode    => '0700',
+    notify  => Service[$service],
+  }
+
+  transition { "stop ${service} service":
```
rtyler: O_o

I don't think I have ever see the `transition` resource

jhoblitt: It is not a well advertised module due to a long list of gotchas.

jhoblitt: Case in point, I had to spend several hours debugging the core the in order get it working right with an absent file resource...

manifests/systemd.pp
jhoblitt: Hmm.  I do that a lot when resource declarations are adjacent.  I can change it to a metaparam.

rtyler: The reason I think we should shy against it is that the `->` placed like this is pretty subtle; I didn't see it on my first scan.

Consistency also isn't a bad thing either

PR #697, opened 2016-12-30
add jenkins::default_plugins param

### LINKED ISSUE #665: Credentials plugin hard coded

jhoblitt:
mash up of #646 and #675

Allows the version of the credentials plugin to be specified via plugin_hash by removing it from the default_plugins list.  Eg.  (Per @jeffmccune)

    jenkins::default_plugins: []
    jenkins::plugin_hash:
      credentials:
        version: 2.1.5
        digest_string: 7db002e7b053f863e2ce96fb58abb98a9c01b09c
        digest_type: sha1

closes #646
closes #675
resolves #665

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -6,7 +6,7 @@ class jenkins::config {
     fail("Use of private class ${name} by ${caller_module_name}")
   }
 
-  ensure_resource('jenkins::plugin', $::jenkins::params::default_plugins)
+  ensure_resource('jenkins::plugin', $::jenkins::default_plugins)
 
   $config_hash = merge(
     $::jenkins::params::config_hash_defaults,
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -291,7 +291,7 @@ class jenkins(
     include jenkins::service
     validate_array($default_plugins)
     if empty($default_plugins){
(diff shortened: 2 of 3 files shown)


jhoblitt:
FYI @elconas @jeffmccune

jeffmccune:
Awesome, thanks!

jhoblitt:
@jeffmccune Sorry for dropping the ball.  I managed to not "post" several review comments right after GH introduced the new review features.

karume:
@jhoblitt Hi! Is this commit going to be tagged eventually? Currently it seems this is only available on master branch. I wouldn't want to neither point my Puppetfile to a master branch of anything or maintain a fork myself.

I wonder if you don't plan to tag anymore? because the latest tag available is from Aug 2016! Thank you

PR #707, opened 2017-01-11
fix acceptance tests on Debian

jhoblitt:


jhoblitt:
I was gonna add some more commits to that. ;)

PR #713, opened 2017-01-16
assorted gem updates, cleanups, and/or removals

jhoblitt:


jhoblitt:
@rtyler Are you still using the blimpy configs?  This PR has a commit that removes them.

rtyler:
@jhoblitt nah, let's burninate them

jhoblitt:
ack - just need to figure out why rubocop is blowing up on puppet 3.8... or we just drop puppet 3.x...

jhoblitt:
@rtyler The rubocop problem appears to be caused by puppet 3.x monkey patching YAML... instead of downgrading rubocop I did some `Rakefile` hackery to keep it skip running it against puppet 3.

PR #723, opened 2017-01-27
Allow specification of additional JAVA_ARGS

cincodenada:
I just upgraded my jenkins module (from 1.4.0), and in upgrading things, ran into the fact that the Jenkins module now manages the INI settings in /etc/default/jenkins (in Debian). I had been setting a few options of my own with `inifile`, which Jenkins' management overwrote.

In the end I decided to solve it by implementing an option to specify additional `JAVA_ARGS` to the module. In doing so, I moved a little bit more of the logic into `config.pp`, in order to allow adding custom JAVA_ARGS.

I wasn't able to get a test environment up, so this PR is partially to get acceptance tests run as well.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -8,8 +8,22 @@ class jenkins::config {
 
   ensure_resource('jenkins::plugin', $::jenkins::default_plugins)
 
+  $java_args = concat(
+    $::jenkins::params::_default_java_args,
+    $::jenkins::java_args
+  )
+
+  $config_defaults = {
+    "${::jenkins::params::config_prefix}JAVA_ARGS" => {
+      value => join($java_args, ' ')
+    },
+    "${::jenkins::params::config_prefix}AJP_PORT" => {
+      value => '-1'
+    }
+  }
+
   $config_hash = merge(
-    $::jenkins::params::config_hash_defaults,
+    $config_defaults,
     $::jenkins::config_hash
   )
   create_resources('jenkins::sysconfig', $config_hash)
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -307,6 +307,7 @@ class jenkins(
   $group              = $::jenkins::params::group,
   $default_plugins    = $::jenkins::params::default_plugins,
   $purge_plugins      = $::jenkins::params::purge_plugins,
+  $java_args          = [],
 ) inherits jenkins::params {
 
   validate_string($version)
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -25,7 +25,7 @@ class jenkins::params {
   $user         = 'jenkins'
   $manage_group = true
   $group        = 'jenkins'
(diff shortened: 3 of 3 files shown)


cincodenada:
Hmm, looks like some relevant tests did indeed fail - I'll take a look at those this week, and maybe take another run at getting my own test environment up and running.

manifests/init.pp
```diff
@@ -307,6 +307,7 @@
   $group              = $::jenkins::params::group,
   $default_plugins    = $::jenkins::params::default_plugins,
   $purge_plugins      = $::jenkins::params::purge_plugins,
+  $java_args          = [],
```
jhoblitt: The default value should come from `::jenkins::params`.

manifests/params.pp
```diff
@@ -25,7 +25,7 @@
   $user         = 'jenkins'
   $manage_group = true
   $group        = 'jenkins'
-  $_java_args   = '-Djava.awt.headless=true -Djenkins.install.runSetupWizard=false'
+  $_default_java_args = ['-Djava.awt.headless=true','-Djenkins.install.runSetupWizard=false']
```
jhoblitt: `jenkins::sysconfig` doesn't work with array values.

manifests/config.pp
```diff
@@ -8,8 +8,22 @@
...
   ensure_resource('jenkins::plugin', $::jenkins::default_plugins)
 
+  $java_args = concat(
+    $::jenkins::params::_default_java_args,
+    $::jenkins::java_args
+  )
+
+  $config_defaults = {
+    "${::jenkins::params::config_prefix}JAVA_ARGS" => {
+      value => join($java_args, ' ')
```
jhoblitt: `jenkins::sysconfig` doesn't work with array values.

manifests/init.pp
jhoblitt: Needs docs, input validation, and test coverage.

manifests/config.pp
```diff
@@ -8,8 +8,22 @@
...
+    },
+    "${::jenkins::params::config_prefix}AJP_PORT" => {
+      value => '-1'
```
jhoblitt: I really don't like to see default values hard-coded outside of the params class. Someday, hopefully soon, we'll be able to replace the params pattern with data-in-modules.

jhoblitt:
I like this feature, and want to move away from the problems with `config_hash/jenkins::sysconfig`, but it is going to be tricky to get the implementation right.

cincodenada:
Thanks for the feedback. As mentioned, this was my problemsolving for my own needs, and PR'd it to see if it was useful enough to round out with tests and closer attention. I'll address your comments and work on tests, thanks!

And yeah, any more substantial changes are going to be pretty tricky, so I think I can shape this up to be a reasonable solution within the existing setup.

bastelfreak:

PR #724, opened 2017-02-02
Let bash expand Java arguments

bambriy:
When multiple $JAVA_ARGS are passed in quotation marks, bash passes entire string as a single argument to java command. To make sure options are digested one by one, java options should not be enclosed in quotes.


Zophar78:
+1... Could we add it to 1.8 milestone?

bambriy:
Builds failed, and not due to the code

PR #458, opened 2016-01-06
Fixes apt module deprecation warnings

vStone:
This also closes pr #444 and some minor puppet-lint warnings.

PR #444, opened 2015-12-03
Feature/fix deprecation warnings

developerinlondon:
currently when using this module we get deprecation warnings from the apt module. 
`
 Warning: Scope(Apt::Source[jenkins]): $include_src is deprecated and will be removed in the next major release, please use $include => { 'src' => false } instead
Warning: Scope(Apt::Source[jenkins]): $key_source is deprecated and will be removed in the next major release, please use $key => { 'source' => http://pkg.jenkins-ci.org/debian/jenkins-ci.org.key } instead.
Warning: Scope(Apt::Key[Add key: 150FDE3F7787E7D11EF4E12A9B7D32F2D50582E6 from Apt::Source jenkins]): $key_source is deprecated and will be removed in the next major release. Please use $source instead.
`
this version fixes the warnings.

PR #733, opened 2017-02-24
Debian repo deprecation warnings

shaunbugler:
The apt module changed the required format in version 2.1, we now get warnings about the old parameters. This merge just updates the repo debian class to bring it inline with the apt module requirements


baurmatt:
Duplicate to #458 #444 #359

Issue #445, opened 2015-12-09
Plugins specified by version number are not updated


alan-schwarzenberger:
The commit at cec172003a66168451e2039cca19a57c2985ae05 changed from wget to archive::download to fetch plugins. This has not implemented correctly fetching plugins by version number.

Take for example https://wiki.jenkins-ci.org/display/JENKINS/Multiple+SCMs+Plugin which we download once say at version 0.3. What we actually have downloaded is the file 'multiple-scms.hpi'.
Given the following plugin code
```
jenkins::plugin { 'multiple-scms':
  version => '0.5',
}
```
Subsequent puppet runs construct a download URL http://updates.jenkins-ci.org/download/plugins/multiple-scms/0.5/multiple-scms.hpi, but uses archive::download where the relevant bit of code is https://github.com/camptocamp/puppet-archive/blob/master/manifests/download.pp#L161 
```
exec {"download archive ${name} and check sum":
  command     => "curl ${proxy_option} -s -S ${insecure_arg} ${redirect_arg} -o ${src_target}/${name} '${url}'",
  creates     => "${src_target}/${name}",
```
And because the file exists (remember there is no version number in the filename) we will not download it again.

One way around it is to specify checksums, but the Jenkins update site does not seem to provide checksums (eg *.md5), so there is no easy way to do that as far as I can see. I don't want to manually download and calculate the checksum in order to update my puppet code. Did I miss something there?

I can see a possible approach which would be to parse "Implementation-Version: 0.3" from META-INF/MANIFEST.MF from inside the hpi file. We could perhaps cheat by reading it from plugins/multiple-scms/META-INF/MANIFEST.MF instead which would be cheaper. If that version did not match the requested version then we could delete the relevant hpi file which would force a re-download, getting the dependency order correct so that it all happens in a single puppet run.

Thoughts on this approach before looking at coding it?

alan-schwarzenberger:
Realised that there is already a jenkins_plugins fact
```
[...]

PR #734, opened 2017-02-28
feature/remove jenkins plugin before downloading it

pouete:
## Why ?
This Pull request assess a problem when upgrading jenkins plugins : The module correctly check if the version is correct and if not, runs into the installation process.
However. as puppet-archive is used and is idempotent, it does not change the version of the so called plugin if it is already installed.

## How ?
The pull request is fairly simple : remove the previous file using exec.
If we use the file puppet method, 
```
file{"${::jenkins::plugin_dir}/${plugin}":
    ensure => absent
}
 
```
the file is already taken care of, and this leads to an error.

### DIFF
--- a/manifests/plugin.pp
+++ b/manifests/plugin.pp
@@ -174,6 +174,9 @@ define jenkins::plugin(
       $checksum_type   = undef
     }
 
+    exec{"force ${plugin}-${version}":
+      command => "/bin/rm -rf ${::jenkins::plugin_dir}/${plugin}"
+    }->
     archive { $plugin:
       source          => $download_url,
       path            => "${::jenkins::plugin_dir}/${plugin}",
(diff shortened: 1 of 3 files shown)


SpencerMalone:
This definitely needs a merge, if possible. Tested it, and it works well. Here's an issue from 2015 describing this very behavior that never got followed up on: https://github.com/jenkinsci/puppet-jenkins/issues/445

jhoblitt:
This seems reasonable but needs rspec coverage before it can be merged.

pouete:
M, looks like the none passing tests are related to a problem with puppet-java ( puppet 4 <-> puppet 3.8 incompatibililty ?

jhoblitt:
Thank you for contributing.  Rebased and merged as #792

PR #743, opened 2017-04-06
Initial FreeBSD support

olevole:
I want to improve module and add support for FreeBSD platform. I plan to implement it in small commits 
chunk/steps and this is the first of them. All changes are tested on FreeBSD 11/12 (aka HEAD)

- add cases for FreeBSD-specific params
- expand exec path to /usr/local/bin which is the default path for 3rd-party software on FreeBSD (and probaly other: HP-UX, Solaris)


olevole:
It will be very good to push https://github.com/jenkinsci/puppet-jenkins/pull/304 into master:  this will greatly facilitate my next commits.

Issue #739, opened 2017-03-17
Swarm 3.3 download URL doesn't exist


mmarseglia:
in manifests/slave.pp the client jar file name is defined as:
`$client_jar = "swarm-client-${version}-jar-with-dependencies.jar"`

There is no swarm client 3.3 with that name at https://repo.jenkins-ci.org/releases/org/jenkins-ci/plugins/swarm-client/3.3/
```
../
swarm-client-3.3-javadoc.jar        10-Feb-2017 08:20  80.47 KB
swarm-client-3.3-javadoc.jar.md5    10-Feb-2017 08:20  32 bytes
swarm-client-3.3-javadoc.jar.sha1   10-Feb-2017 08:20  40 bytes
swarm-client-3.3-sources.jar        10-Feb-2017 08:20  17.78 KB
swarm-client-3.3-sources.jar.md5    10-Feb-2017 08:20  32 bytes
swarm-client-3.3-sources.jar.sha1   10-Feb-2017 08:20  40 bytes
swarm-client-3.3.jar                10-Feb-2017 08:20  1.50 MB
swarm-client-3.3.jar.md5            10-Feb-2017 08:20  32 bytes
swarm-client-3.3.jar.sha1           10-Feb-2017 08:20  40 bytes
swarm-client-3.3.pom                10-Feb-2017 08:20  3.89 KB
swarm-client-3.3.pom.md5            10-Feb-2017 08:20  32 bytes
swarm-client-3.3.pom.sha1           10-Feb-2017 08:20  40 bytes
```

SpencerMalone:
It seems that without announcing it in the changelogs, they removed the dependency-less jar and renamed the jar-with-dependencies to swarm-client-3.3.jar, so hopefully we just have to support that new URL name if version >= 3.3. See: https://issues.jenkins-ci.org/browse/JENKINS-42138

jantman:
Yeah, that's what happened. I suppose the right way would be to change ``jenkins::slave::client_jar`` based on ``$version``, but for future-proofing, it would probably be good if that was exposed as a param as well.

jantman:
It looks like #746 includes a code fix for this, but no test updates.

jhoblitt:
This has indeed been fixed on master for sometime, so I'm going to close this ticket.

PR #746, opened 2017-04-11
Deal with swarm filename changing in repos.jenkins-ci.org

bashou:
Added versioncmp to change dynamically name of file swarm-client* downloaded when managed (or not)

Resolving issue #739 too


pgassmann:
@bashou Can you update the spec test accordingly?

jhoblitt:
This looks good to me, just needs rspec coverage.

PR #750, opened 2017-05-04
avoid getting plainText from null passphrase (in case key was created with empty passphrase)

kristofvb:


jhoblitt:
There are no commits on this PR so I'm going to close it.  Please feel free to open a new PR.

PR #751, opened 2017-05-05
Added legacy functionality when using the `-s` flag in the jenkins cli

darioatbashton:
Latest jenkins release is not accepting anymore the `-s` flag to use the cli as admin with an ssh key using the http api, we have to add the `-remoting` option to use this functionality.

### DIFF
--- a/lib/puppet_x/jenkins/provider/cli.rb
+++ b/lib/puppet_x/jenkins/provider/cli.rb
@@ -163,7 +163,7 @@ class PuppetX::Jenkins::Provider::Cli < Puppet::Provider
 
     auth_cmd = nil
     unless ssh_private_key.nil?
-      auth_cmd = base_cmd + ['-i', ssh_private_key] + [command]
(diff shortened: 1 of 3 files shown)


jhoblitt:
Could you rebase on current master?  The puppet-lint warnings should be fixed.

jhoblitt:
Actually, this isn't go to be backwards compatible-enough to be a workable fix.

mouellet:
You would also have to add the `-remoting` option at https://github.com/jenkinsci/puppet-jenkins/blob/master/lib/puppet_x/jenkins/provider/cli.rb#L166

darioatbashton:
@jhoblitt I rebased my master. I'm using this fix and it works ok for me so far, why do you think is not going to be compatible?

rtyler:
@jhoblitt The `-remoting` flag is now used for all currently supported versions of Jenkins by the project. From my perspective that is sufficient bar to merge something like this for the next release.

PR #752, opened 2017-05-06
Fix https://github.com/jenkinsci/puppet-jenkins/issues/749

### LINKED ISSUE #749: puppet_helper.groovy uses deprecated remoting mode
My puppet config uses the bootstrap script as described in [0]. However, it appears to have been broken by the deprecation of remoting mode [1]. When puppet runs I now get:
```
java -jar /usr/share/jenkins/jenkins-cli.jar -s http://127.0.0.1:8080 -http groovy /usr/share/jenkins/puppet_helper.groovy create_or_update_user admin me@me.com  admin <SSH KEY>

ERROR: This command is requesting the deprecated -remoting mode. See https://jenkins.io/redirect/cli-command-requires-channel
```

It looks as if we now need to use the ssh or http authentication methods in the `create_or_update_user` function of `puppet_helper.groovy`.

[0] https://github.com/jenkinsci/puppet-jenkins#cli-helper
[1] https://github.com/jenkinsci/jenkins/pull/2795

### LINKED ISSUE #753: Commit breaks tests
This commit https://github.com/jenkinsci/puppet-jenkins/commit/a2e21e8733909799bb5285230234892858712a9a adds test but without puppet_helper.groovy implementation breaks beaker tests as there is no implementation

elconas:
This fixes #749 by adding support for the CLI introduced in jenkins 2.54 / 2.46.2 LTS 

https://issues.jenkins-ci.org/browse/JENKINS-41745

Also fixes #753 and some other test failues (like lint etc.)

### DIFF
--- a/lib/puppet_x/jenkins/provider/cli.rb
+++ b/lib/puppet_x/jenkins/provider/cli.rb
@@ -146,7 +146,12 @@ class PuppetX::Jenkins::Provider::Cli < Puppet::Provider
     tmp.write input
     tmp.flush
     options[:stdinfile] = tmp.path
-    FileUtils.chown 'jenkins', 'jenkins', tmp.path if tmpfile_as_param and File.exists?(tmp.path)
+    begin
+      Etc.getpwnam('jenkins')
+      FileUtils.chown 'jenkins', 'jenkins', tmp.path if tmpfile_as_param and File.exists?(tmp.path)
+    rescue
+      FileUtils.chmod 0644, tmp.path if tmpfile_as_param and File.exists?(tmp.path)
+    end
     result = execute_with_retry(command, options, cli_pre_cmd)
     tmp.close
     tmp.unlink
--- a/lib/puppet_x/jenkins/provider/cli.rb
+++ b/lib/puppet_x/jenkins/provider/cli.rb
@@ -150,7 +150,7 @@ class PuppetX::Jenkins::Provider::Cli < Puppet::Provider
       Etc.getpwnam('jenkins')
       FileUtils.chown 'jenkins', 'jenkins', tmp.path if tmpfile_as_param and File.exists?(tmp.path)
     rescue
-      FileUtils.chmod 0644, tmp.path if tmpfile_as_param and File.exists?(tmp.path)
+      FileUtils.chmod 0o644, tmp.path if tmpfile_as_param and File.exists?(tmp.path)
     end
     result = execute_with_retry(command, options, cli_pre_cmd)
     tmp.close
--- a/lib/puppet_x/jenkins/provider/cli.rb
+++ b/lib/puppet_x/jenkins/provider/cli.rb
@@ -96,21 +96,32 @@ class PuppetX::Jenkins::Provider::Cli < Puppet::Provider
     self.class.cli(*args)
   end
 
-  def self.clihelper(command, options = nil)
-    catalog = options.nil? ? nil : options[:catalog]
+  def self.clihelper(command, options = {})
+    catalog = options.key?(:catalog) ? options[:catalog] : nil
     config = PuppetX::Jenkins::Config.new(catalog)
 
     puppet_helper = config[:puppet_helper]
+    cli_remoting_free = config[:cli_remoting_free]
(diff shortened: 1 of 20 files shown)


oc243:
Thanks @elconas this looks great.

One question for you. I can't quite get this PR to work with the experimental `jenkins_credentials` providers. It looks like your `-auth 'login:pass'` trick needs to be applied to the command at [0], else you hit the same issue of using the remoting API. Do you agree? Would you be happy to make the changes? If not I can have a go.

[0] https://github.com/jenkinsci/puppet-jenkins/blob/56f05d26aff21c35f7900a5843f9af36be0a2a75/lib/puppet_x/jenkins/provider/cli.rb#L153

oc243:
Thanks for the @elconas. I've used your patch and it is working on my setup :)

elconas:
Merged Master, rebased and changed author.

themasterchef:
We have used your patch on our setup at work and can also confirm that it works.

We prefer this over the similar pull request as it's a bit smaller in scope and doesn't make as many modifications to how users work, which might help this one get merged more quickly.

Whichever solution is chosen, a merge to upstream is *fairly urgent* because any Jenkins 2.46.2+ installation that is set up with this module will be *totally* hosed by the remoting-free CLI changes - in our setup Jenkins couldn't even be bootstrapped with credentials.

lib/puppet_x/jenkins/provider/cli.rb
```diff
@@ -122,22 +133,31 @@ def self.cli(command, options = {})
...
+
     Puppet.debug("#{sname} stdin:\n#{input}")
 
     # a tempfile block arg is not used to simplify mock testing :/
     tmp = Tempfile.open(sname)
     tmp.write input
     tmp.flush
     options[:stdinfile] = tmp.path
-    result = execute_with_retry(command, options)
+    FileUtils.chown 'jenkins', 'jenkins', tmp.path if tmpfile_as_param and File.exists?(tmp.path)
```
rtyler: This user/group isn't guaranteed to exist in every installation, though they are guaranteed to exist with the native packages Jenkins distributes.

This line seems a bit unrelated to everything else that's changed in this file, please add a followup commit that explains what this is for and perhaps makes it more parameterized.

stefanandres:
Please also fix https://github.com/elconas/puppet-jenkins/blob/fix_749/manifests/repo/debian.pp#L18 - this should just be a hash like
```
key => {
  id => ...
  source => ...
}
```
(See https://github.com/puppetlabs/puppetlabs-apt/blob/4.1.0/manifests/source.pp#L10, in the current master branch it is fine -> https://github.com/jenkinsci/puppet-jenkins/blob/master/manifests/repo/debian.pp#L20)

lib/puppet_x/jenkins/provider/cli.rb
elconas: We use umask 027. Then this file is created by root and not readable by jenkins.

elconas:
@rtyler Actually I removed the -remoting flag in this pull request, as I think this module is used in different setups:

* OLD Jenkins (pre 2.54 / 2.46.2) - aka remoting free = false
  * unsecured - no -remoting, as not available
  * secured - no -remoting, as not available
* NEW Jenkins (2.54++/2.46.2++) - aka remoting free = true
  * unsecured - no use -auth (-remoting not required)
  * secured - use -auth user:pass or -auth @filename to authenticate

So there is no need for -remoting anymore as the implementation in this pull request only uses remoting free setups. So remoting can be disabled at all (recommended by Jenkins anyway).

elconas:
@stefanandres I backported master, so the fix is included.

elconas:
I disabled puppet 3 in .travis.yml, as master contains Puppet4 native code (Puppet Data Types), causing compile failures.

robinbowes:
> master contains Puppet4 native code (Puppet Data Types), causing compile failures.

This is disappointing.

I was hoping there would be a 1.8.0 release before puppet 3 support was abandoned (presumably with a 2.0.0 release).

Where's the puppet4 native code?

PR #754, opened 2017-05-10
Fix all require statements for files in puppet_x

vStone:
See: https://tickets.puppetlabs.com/browse/SERVER-973 for more
information


vStone:
I'm looking to fix the PUPPET_X warnings

PR #756, opened 2017-05-10
Correct all arrow alignments using puppet-lint

vStone:


jhoblitt:
Doh! I just opened a PR for this too.  I will merge this when travis finishes.

PR #759, opened 2017-05-10
Feature/job replace parameter

vStone:
This is a re-implementation of #500. I also threw in tests in there and some indentation changes also slipped in there.

### DIFF
--- a/lib/puppet/type/jenkins_job.rb
+++ b/lib/puppet/type/jenkins_job.rb
@@ -29,7 +29,10 @@ Puppet::X::Jenkins::Type::Cli.newtype(:jenkins_job) do
       elsif newvalue == :absent
         'removed'
       else
-        if Puppet[:show_diff] && resource[:show_diff]
+        if @resource[:replace] == false
+          return 'left unchanged'
+        end
+        if Puppet[:show_diff] and resource[:show_diff]
           # XXX this really should be turned into a helper method and submitted
           # to # core puppet
           Tempfile.open('puppet-file') do |d1|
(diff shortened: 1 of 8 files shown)


jhoblitt:
I'd rather not add any new functionality to `jenkins::job` that isn't present in the native type as it ends up adding yet another hurdle to changing the defined types to being wrappers around the native types.  Are you willing to fix up the native type as well?

vStone:
@jhoblitt I was gonna adjust the acceptance tests as well for the native types since I don't use those myself yet but it seems that no native types are being tested there.

vStone:
@jhoblitt bump: rebased + tests are green!

vStone:
This pr already includes support for the old jenkins::job and the new jenkins_job.

vStone:
@jhoblitt anything that is stopping you from merging this one?

manifests/job.pp
```diff
@@ -27,6 +27,9 @@
 #   difftool = '/usr/bin/diff -b-q'
 #     Provide a command to execute to compare Jenkins job files
 #
+#   replace = 'true'
+#     Wether or not to replace the job if it already exists.
```
alexjfisher: https://brians.wsu.edu/2016/05/19/weather-wether-whether/ ;)

manifests/job/present.pp
```diff
@@ -18,12 +18,16 @@
 #   enabled
 #     deprecated parameter (will have no effect if set)
 #
+#   replace = 'true'
+#     Wether or not to replace the job if it already exists.
```
alexjfisher: https://brians.wsu.edu/2016/05/19/weather-wether-whether/

lib/puppet/type/jenkins_job.rb
```diff
@@ -64,6 +67,11 @@ def change_to_s(currentvalue, newvalue)
     defaultto true
   end
 
+  newproperty(:replace, :boolean => true, :parent => Puppet::Property::Boolean) do
```
alexjfisher: Puppet::Property::Boolean is totally broken. (Setting something that is true to false just doesn’t work) But... should this actually be a parameter instead of a property?

jhoblitt:
This seems reasonable.  Thank you for being persistent.

PR #760, opened 2017-05-10
Avoid getting plainText from null passphrase 

vStone:
in case key was created with empty passphrase


jhoblitt:
It would be nice to have some sort of test coverage but those seems harmless.

PR #780, opened 2017-07-07
Bring start-slave script on MacOS inline with other OSes

### LINKED ISSUE #779: Slave startup arguments under MacOS
From what I can see the startup scripts and config don't actually pass in any arguments to the start script under MacOS.

The start-slave.sh script passes $JENKINS_SLAVE_ARGS to the start command but that variable isn't set anywhere at all.

I am going to attempt to fix this.

rendhalver:
The start-slave script was not using the env vars properly
This generates the start up options from the env vars in jenkins-slave

Fixes #779

### DIFF
--- a/templates/start-slave.sh.erb
+++ b/templates/start-slave.sh.erb
@@ -1,4 +1,68 @@
 #!/usr/bin/env bash
 
-source <%= @defaults_location %>/jenkins-slave 
-java -jar <%= @slave_home %>/<%= @client_jar %> $JENKINS_SLAVE_ARGS
+source <%= @defaults_location %>/jenkins-slave
+
+# mandatory input vars
+[[ -x "$JAVA" ]] || (fail "$JAVA is not executable")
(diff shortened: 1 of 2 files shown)


templates/start-slave.sh.erb
```diff
@@ -1,4 +1,68 @@
 #!/usr/bin/env bash
 
-source <%= @defaults_location %>/jenkins-slave 
-java -jar <%= @slave_home %>/<%= @client_jar %> $JENKINS_SLAVE_ARGS
+source <%= @defaults_location %>/jenkins-slave
+
+# mandatory input vars
```
jhoblitt: Instead of cut'n'pasting the the entire `jenkins-slave-run` script, how about installing it and call it from this script?

rendhalver: Sounds like a good plan.
I think I did it this way because it wasn't an exact copy of the linux script.
I will see if I can reuse the linux script.

Issue #767, opened 2017-06-02
Prevents upgrade of puppetlabs-java module to 2.0.0


leitmedium:
Currently puppet-jenkins allows puppetlabs-java only to be < 2.0.0, but 2.0.0 has been released. Upgrading puppetlabs-java to the latest version is therefore not possible at the moment.

jayniz:
I, too, wonder!

mmoll:
The same is true for puppetlabs-apt... what are the plans for the next release? Especially related to #772

cristianbalan:
Any news here? :(

jhoblitt:
PR requested.

mmoll:
see GH-801

PR #801, opened 2017-09-02
allow newer puppetlabs-java version

mmoll:
closes GH-767


mmoll:
AFAICT, the 2.x versions are 100% API compatible

PR #802, opened 2017-09-02
allow newer puppetlabs-apt versions

mmoll:


mmoll:
AFAICT, these versions are 100% API compatible

PR #803, opened 2017-09-02
allow newer puppet-archive versions

mmoll:


mmoll:
AFAICT, 2.x is 100% API compatible

PR #809, opened 2017-09-22
Adding PAM as a user realm

TheKangaroo:
Like #606 for LDAP, this PR adds the pam authentication for Jenkins.
The configuration of pam Auth works fine for me. 
I still have trouble with Jenkins only supporting one security realm provider (https://issues.jenkins-ci.org/browse/JENKINS-15063).
Therefore as soon as I configure PAM auth with puppet, the puppet Jenkins configuration user (who is in Jenkins own user database) stops working and there seems to be no way, to automate the deployment of Jenkins with pam.
The only workaround I can think of would be a user in the Linux system with the exact same username and password as the Jenkins admin+initialAdminPassword.
But I'm not sure how to implement this right now.
Any ideas or suggestions on that?

### DIFF
--- a/files/puppet_helper.groovy
+++ b/files/puppet_helper.groovy
@@ -24,6 +24,8 @@ import groovy.transform.InheritConstructors
 import hudson.model.*
 import hudson.plugins.sshslaves.*
 import hudson.util.*
+import hudson.security.PAMSecurityRealm
+import hudson.security.HudsonPrivateSecurityRealm
 import jenkins.model.*
 import jenkins.security.*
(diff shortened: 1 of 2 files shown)


jhoblitt:
@TheKangaroo It looks like there are at least syntax problems with the groovy.

@rtyler I sort of feel like `security.pp` needs to die in a fire.  Do you want to accept new non-native type changes?  I'd prefer that this go into `jenkins_security_realm` type.

bastelfreak:

PR #810, opened 2017-10-17
Fixed issue where CLI helper keeps getting a 503 from the server when…

wiccan2:
Fixed 503 error when using CLI helper in our production environment.


jhoblitt:
We need information about the environment, what caused the problem and why this is a fix.  I've never  encountered a 503 due to URL construction.

wiccan2:
@jhoblit I have now reverted this in our environment and we are no longer getting the issue.

Please feel free to reject this pull request as I don't think it is needed.

jhoblitt:
Roger that.

PR #811, opened 2017-10-27
Fixing incorrect port used by CLI when non default specified.

steve-stonehouse:
Using `jenkins::port` in Hiera did not change the port the CLI uses.


jhoblitt:
Unfortunately, the semantics here are confusing and `jenkins::port` is a newish addition on top of an existing mess.  I think I'd rather the function be updated to to use `::jenkins::port` as the highest weight value.  What do you think @rtyler ?

bastelfreak:

PR #819, opened 2017-11-23
Make all archive options accessible for plugin type

stefanbethke:
At our organization, we are downloading plugins from our own caching
mirror, but we can't access our internal mirror through our proxy
server, and we do need to authenticate to this server.

With this change, archive (and in turn, curl) will pick up the proxy
configuration from environment variables or ~/.curlrc, including the
no_proxy setting we need. When users desire to override those, they can
still specify proxy_server directly.

For authenticating with the download server, we can pass username and
password to archive.

In our private wrapper module, we set default for the jenkins::plugin
type:
```
Jenkins::Plugin {
  archive_options => |
    username => $artifactory_username,
    password => $artifactory_password,
}
```

This works because we define the defaults and the actual resources
inside the same class. Setting defaults for types is guaranteed by
Puppet only for resources declared in the same class, so we cannot
set defaults on the archive type.

If full backwards compatibility is desired, the parameter declaration
could be changed to:
```
  Hash $archive_options = { 'proxy_server' => $::jenkins::proxy::url },
```

### DIFF
--- a/manifests/plugin.pp
+++ b/manifests/plugin.pp
@@ -26,6 +26,7 @@ define jenkins::plugin(
   Boolean $enabled                  = true,
   String $digest_type               = 'sha1',
   Boolean $pin                      = false,
+  Hash $archive_options             = {},
   # no worky
   Any $timeout                      = undef,
   # deprecated
@@ -173,11 +174,11 @@ define jenkins::plugin(
       checksum_verify => $checksum_verify,
       checksum        => $checksum,
       checksum_type   => $checksum_type,
-      proxy_server    => $::jenkins::proxy::url,
       cleanup         => false,
       extract         => false,
       require         => File[$::jenkins::plugin_dir],
       notify          => Service['jenkins'],
+      *               => $archive_options,
     }
     $archive_require = Archive[$plugin]
   } else {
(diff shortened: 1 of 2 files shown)


jhoblitt:
I'm  fine with being able to override the global proxy settings on a per `jenkins::plugin` basis but I'm really not keen on exposing the semantics of the `archive` type in the public API.  There has been already been a lot of pain in this module when trying to do backwards compatible refactoring.  I think it would make sense to eventually replace `$::jenkins::proxy_host` and `$::jenkins::proxy_port` with a single param, perhaps `$::jenkins::proxy` or `proxy_url`.  So it would make sense to add only a `proxy-ish` param to this type and continues to default to `$::jenkins::proxy::url`.

stefanbethke:
I understand your converns, but how do you propose that we ensure that login and no_proxy information for our mirror repository are passed to archive (and eventually curl)?

jhoblitt:
I believe the standard is `schema://username:password@hostname:port` isn't?

stefanbethke:
Specifying username and password as part of the URL might be one option, true.

We've now reconfigured our internal systems so that we can not configure a proxy for UpdateCenter, and do not need authentication to access our mirror anymore, so we don't need this change anymore.

I'd like to point out, however,  that the abstraction of voxpopuli/archive and curl is leaky when it comes to needing `no_proxy` env var or the `$no_proxy_list` parameter: when you specify `$proxy_server` for archive, archive will call curl with the explicit `--proxy` parameter, disabling all parsing of env variables and `~/.curlrc`, effectively disabling any proxy exception list. This is ultimately a defect in archive, but jenkins triggers the bug by passing the parameter.

jhoblitt:
I'm going to go ahead and close this out.

PR #820, opened 2017-11-24
Add missing dependency to the gitlab plugin (credentials) test

vStone:
Looks like it might be a dependency issue since gitlab-plugin depends on new versions of everything.


jhoblitt:
:+1:

PR #821, opened 2017-11-28
Run apt_update when defining debian repos

beezly:
As per the puppetlabs/apt documentation, there should be a `notify => Exec['apt_update']` when defining repos using apt::sources, otherwise there is no guarantee that the packages in the new repo will be available the next time Apt tries to install a package.


jhoblitt:
Sounds reasonable.  Could you add unit test coverage of the resource relations somewhere like `spec/classes/jenkins_repo_debian_spec.rb`?

beezly:
@jhoblitt cheers. I've added another commit with a check for the notifies attribute on the jenkins apt::source resource.

PR #822, opened 2017-11-28
bump dependency on camptocamp/systemd

costela:
Just a small dependency bump. Can't find any changes in the systemd 1.* series that might conflict with the use in this module.


jhoblitt:
Seems reasonable to me. :+1:

PR #761, opened 2017-05-11
Add bootstrapping support for jenkins users (secure installations).

elconas:
This is another version of https://github.com/jenkinsci/puppet-jenkins/pull/752 that includes support for bootstrapping users via init.groovy.d as this was requested by comments in the issue thread linked to https://github.com/jenkinsci/puppet-jenkins/pull/752 -> https://github.com/jenkinsci/puppet-jenkins/issues/749 ... 

Other issues that may interest this pull are: 
  - https://github.com/jenkinsci/puppet-jenkins/issues/749
  - https://github.com/jenkinsci/puppet-jenkins/issues/671
  - https://github.com/jenkinsci/puppet-jenkins/issues/684
  -  .. maybe others.

PR #824, opened 2017-11-30
rebased pr 761+ssh pubkey support

vStone:
In an attempt to get #761 merged, I've rebased it and added support for a ssh pubkey configuration in the bootstrap user.


jhoblitt:
I think it would be reasonable to merge management of `init.groovy.d` at this point but not the built in templating.  I don't think it offers functionality over the existing native types and jenkins "configuration as code" is supposed to be shipping soon.

vStone:
I'm not too happy with the code either but it does work for some use-cases with additional authentication/authorization setup. I'm inclined to close it but on the other hand I rather keep it here if somebody is looking for a workaround (for now.)

PR #832, opened 2018-01-31
drop legacy puppet version from testmatrix

bastelfreak:
There is no need to test against two Puppet 4.X versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,6 @@ rvm:
 script:
   - "bundle exec rake SPEC_OPTS='--format documentation --order random'"
 env:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -8,7 +8,6 @@ rvm:
 script:
   - "bundle exec rake SPEC_OPTS='--format documentation --order random'"
 env:
-  - PUPPET_VERSION="~> 4.9.0"
```
alexjfisher: Legacy *ruby* version??

bastelfreak: puppet, ruby... all the same :D

PR #835, opened 2018-02-05
Fixes password quoting for jenkins.run template

tthayer:
Passwords with special characters don't work.
Changing quoting from single to double in the template allows the password to be passed properly. This has been tested on RHEL/Centos 7.


jhoblitt:
The acceptance test failure doesn't appear to be related so I'm going to merge this.

tthayer:
Great! Thank you.

PR #348, opened 2015-07-17
Added tunnel arg to jenkins::slave, to support this arg for the swarm…

fipp:
Support for the tunnel argument added to the Jenkins Swarm Plugin in version 1.23.

PR #840, opened 2018-02-07
Feature/slave tunnel

jhooyberghs:
Added tests for existing open PR "Add tunnel flag for slaves" #796


jhooyberghs:
Should also close #348

manifests/slave.pp
```diff
@@ -27,6 +27,11 @@
 # [*executors*]
 #   Number of executors for this slave.  (How many jenkins jobs can run simultaneously on this host.)
 #
+# [*tunnel*]
```
ghoneycutt: This looks great! If we used `tunnel_host` and `tunnel_port` it might be easier to validate the inputs and then we could compose `$tunnel` inside the class.

jhooyberghs: We could, however then we would also need to verify if one of those have been given or not to see if you have to print something because
"HOSTNAME:"
":PORT"
"HOSTNAME:PORT"
are all valid, while  ":" is not 
(source https://github.com/jenkinsci/jenkins/blob/master/war/src/main/webapp/help/system-config/master-slave/jnlp-tunnel.html)

ghoneycutt: Gotcha, that part wasn't clear to me reading the doc string. Thanks!

vStone: We could catch this using the data types. (regex magic maybe)

jhooyberghs: Like this @vStone?

vStone: <3

ghoneycutt:
Created https://github.com/jhooyberghs/puppet-jenkins/pull/1 to add some spec tests

jhoblitt:
The acceptance test failure should be "fixed" (marked as pending) on master.

PR #845, opened 2018-02-08
Disable rubocop for okjson.rb and resync from upstream. Fix Rubocop Style/RegexpLiteral

alexjfisher:
okjson came from https://github.com/kr/okjson so I think it was a mistake to try to apply rubocop fixes to this file.

Having reverted the recent rubocop changes, I have resynced the file to the upstream version (2 small fixes), and added okjson.rb to the rubocop excludes.

I've then fixed the remaining rubocop Style/RegexpLiteral violations (both autofixes and manual ones).  This allowed me to create a rubocop_todo.yml and renable running rubocop in the default rake task.


.rubocop.yml
```diff
@@ -13,6 +14,7 @@ AllCops:
     - Rakefile
     - Guardfile
     - Vagrantfile
+    - lib/puppet/jenkins/okjson.rb
```
alexjfisher: @bastelfreak I don't think modulesync_config currently supports adding per module excludes like this?

jhoblitt:
FYI - I have a vague memory that @rtyler (or someone) had modified okjson, so it may not be possible to copy upstream verbatim.

jhoblitt:
Actually, it looks like `#plugins_from_updatecenter` is mangling the JSONP-ness, so it is probably fine.
Double actually, `json` is core these days so a bundled implementation probably isn't needed at all.

jhoblitt:
@rtyler ^ ???

alexjfisher:
The changes from the resync are in a separate commit to the revert of the rubocop related changes.  See https://github.com/voxpupuli/puppet-jenkins/pull/845/commits/a4c825d57b118a532cbf4ccd97cf5a57a3e00722
I think it's just the last 4 closed PRs https://github.com/kr/okjson/pulls?q=is%3Apr+is%3Aclosed

alexjfisher:
Our version didn't pass the upstream testsuite. https://github.com/kr/okjson/blob/master/test

.rubocop.yml
bastelfreak: It doesn't. I try to avoid this and and disable the 'broken' cops in the file itself.

alexjfisher: That would make periodically syncing from the upstream a bit more of a pain I guess.  I suppose we could just disable all applicable cops in 20 or so lines directly at the top of the file?  We'll need to decide when we get to the point we can actually try the modulesync again.

PR #846, opened 2018-02-09
Add feature to disable jenkins-slave service

ArVincentr:
if the swarm plugin is not installed and the master reaches out the slave and push its jar  file, service is not needed.
What has been done? :

Create a boolean $manage_slave_service ( set to true by default ). if set to false, service bits are not executed.

### DIFF
--- a/manifests/slave.pp
+++ b/manifests/slave.pp
@@ -140,6 +140,7 @@ class jenkins::slave (
   Any $java_args                          = undef,
   Any $swarm_client_args                  = undef,
   Boolean $delete_existing_clients        = false,
+  Boolean $manage_slave_service           = true,
 ) inherits jenkins::params {
 
   if versioncmp($version, '3.0') < 0 {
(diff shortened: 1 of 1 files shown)


jhoblitt:
I'm not sure I understand the objective here, if the intent is to use the sshslaves plugin, why would the `jenkins::slave` class be used at all?

ArVincentr:
Hi, 
I understand your point here.  our Jenkins landscapes have been setup manually by different teams in the past, and they are indeed using the sshslave plug-in you mentioned.
Time has come to integrate it in our puppet codebase. The puppet-jenkins module is awesome and we want to use it. the goal is to give more flexibility to the jenkins module and allow us ( and probably other ppl) to migrate slowly to jenkins-swarm plug-in and thus make use of the entire set of functionalities your module provides. the "feature" I'd like to propose here can give some flexibility when it comes to integrate  mixed kind of jenkins slaves.

jhoblitt:
@ArVincent For an `sshslave` host you need at java, a role account, and a working `sshd` configuration.   Disabling virtually of the functionality in a class intended to managed a jenkins swarm client, that still isn't able to provide a working `sshslave` host seems unreasonable to me and would require unit tests that verify all of the unwanted resources aren't appearing in the catalog -- that's just not worth it my opinion to save the effort of adding an `import` and a `user` resource to a role/profile that will be needed to manage `sshd` configuration anyways.  Eg.

```
class { ::mysite::role::jenkinssh:
  include ::java
  user { 'mysshuser": ... }
  ...
}
```

PR #852, opened 2018-02-18
4 more rubocop fixes

alexjfisher:


lib/puppet/provider/jenkins_job/cli.rb
```diff
@@ -78,7 +78,8 @@ def self.get_job(job, catalog = nil)
   # currently unused
   def self.job_enabled(job, catalog = nil)
     raw = clihelper(['job_enabled', job], catalog: catalog)
-    !!(raw =~ %r{true})
+    return true if raw =~ %r{true}
+    false
```
alexjfisher: There's probably a better way that still doesn't use `!!`.  Open to suggestions.

ekohl: https://stackoverflow.com/a/41707707 suggests ```raw =~ %r{true} ? true : false```

jhoblitt: What is the concern here? The meaning of `!!` is non-ambiguous.

alexjfisher: @jhoblitt I don't disagree, and others also share your view. eg See https://github.com/bbatsov/ruby-style-guide/issues/591

Meanwhile, I think @ekohl's suggestion is less ugly than using a control comment.
```
!!(raw =~ %r{true}) # rubocop:disable Style/DoubleNegation
```

I've been trying to get rid of .rubocop_todo.yml so that we can finally do the modulesync required for us to be able to release using the normal Vox Pupuli process.  With this PR we're down to 'just' 196 violations having originally started at more than 3000.  We probably will end up using *some* control comments to disable certain cops, but I've been trying to resist so far.

PR #861, opened 2018-03-28
Add support for GoogleRobotPrivateKeyCredentials

thaiphv:
Extends the `credentials_list_json` method in the `puppet_helpers.groovy` script to support `com.google.jenkins.plugins.credentials.oauth.GoogleRobotPrivateKeyCredentials`.

### DIFF
--- a/files/puppet_helper.groovy
+++ b/files/puppet_helper.groovy
@@ -547,8 +547,19 @@ class Actions {
           }
           break
         case 'com.google.jenkins.plugins.credentials.oauth.GoogleRobotPrivateKeyCredentials':
-          info['account_id'] = cred.getServiceAccountConfig().getAccountId()
-          info['private_key'] = IOUtils.toString(cred.getServiceAccountConfig().getPrivateKey().getEncoded(), "UTF-8")
+          info['json_key'] = null
+          info['email_address'] = null
+          info['p12_key'] = null
+
+          def serviceAccountConfig = cred.getServiceAccountConfig()
+          if (serviceAccountConfig.getClass().getName() == 'com.google.jenkins.plugins.credentials.oauth.JsonServiceAccountConfig') {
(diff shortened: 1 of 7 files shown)


files/puppet_helper.groovy
```diff
@@ -546,6 +546,10 @@ class Actions {
               throw new UnsupportedCredentialsClass("unsupported " + keyStoreSource)
           }
           break
+        case 'com.google.jenkins.plugins.credentials.oauth.GoogleRobotPrivateKeyCredentials':
+          info['account_id'] = cred.getServiceAccountConfig().getAccountId()
+          info['private_key'] = new String(cred.getServiceAccountConfig().getPrivateKey().getEncoded())
```
ekohl: Others use `IOUtils.toString`. As someone not that familiar with groovy, is there a difference?

thaiphv: TBH I'm not sure either but I'm happy to make that change.

ekohl: Maybe some more experienced Jenkins/groovy user could comment on this.

jhoblitt: It looks OK as-is.  A literal string is being used to avoid requiring all of those classes to be loaded.

jhoblitt:
I restarted the travis bulid last night after the first failure but it looks like the last build was only 4 hours ago.  Was the PR updated with a fix?  I want to make sure we're not seeing unstable test behavior.

thaiphv:
@jhoblitt, I hasn't updated it since the first commit. I think I need to update the `credentials_update_json` method as well.

thaiphv:
I've also extended the ``credentials_update_json` method so that we could add credentials of this type.

bastelfreak:
Thanks for the work @thaiphv !

PR #863, opened 2018-04-01
modulesync 1.9.0

bastelfreak:


.travis.yml
```diff
@@ -1,30 +1,49 @@
...
   include:
-    - rvm: '2.1.9'
-      bundler_args: --without development system_tests
-      env: PUPPET_VERSION="~> 4.0" CHECK=test PARALLEL_TEST_PROCESSORS=12
-      script: "bundle exec rake SPEC_OPTS='--format documentation --order random'"
-    - rvm: '2.4'
-      sudo: required
-      services: docker
-      env: BEAKER_set="centos-7-docker"
-      script: travis_retry bundle exec rake beaker
```
bastelfreak: we need to add this to our .sync.yml

bastelfreak: fixed

bastelfreak:
Fixes https://github.com/voxpupuli/puppet-jenkins/issues/866

alexjfisher:
@bastelfreak What did you want to do about the .rubocop_todo.yml?  I ran out of energy! :(

Related, I think we'll need something in modulesync_config to support this line https://github.com/voxpupuli/puppet-jenkins/blob/ac2bdb212b2987f3eda127f0748810f4126a2513/.rubocop.yml#L17

bastelfreak:
@alexjfisher IMO we can just keep it as is for now. The acceptance tests are a different story. I've no idea how to fix them.

manifests/plugin.pp
```diff
@@ -91,7 +91,7 @@
     fail("unsupported plugin extension in source url: ${download_url}")
   }
 
-  $installed_plugins = $::jenkins_plugins ? {
+  $installed_plugins = fact('jenkins_plugins') ? {
```
alexjfisher: How come the `fact` function here, but in params.pp the `$facts` hash?

bastelfreak:
@jhoblitt do you have any idea about the failing tests?

jhoblitt:
The acceptance tests worked on master for me a little while ago.  Perhaps this PR could be split up into smaller pieces?  There is alot going on here and most of the changesets shouldn't be dependent upon each other.  In particular, the `puppet-lint`/`rubocop` auto-fixes should not be coupled with the other change sets as if they are, the auto-fixing was "auto" but not "fix". ;)

PR #868, opened 2018-05-18
Update jenkins-run.erb

sbeaulie:
see [JENKINS-51430](https://issues.jenkins-ci.org/browse/JENKINS-51430)

### DIFF
--- a/templates/jenkins-run.erb
+++ b/templates/jenkins-run.erb
@@ -50,8 +50,6 @@ do
   JENKINS_JAVA_CMD="$candidate"
 done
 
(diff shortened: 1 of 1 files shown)


sbeaulie:
@ekohl The issue is systemd because it parses for spaces and feeds that to the Exec process as arguments, even if said space is in quotes. Thats not how bash works.

@jhoblitt I tested this PR and it turns out it only works for the first option in `$JENKINS_JAVA_OPTIONS` which is not what I am looking for. I'm not too familiar with modifying the scripts for systemd, would it need to be similar to how $PARAMS is filled?

jhoblitt:
This script is pure bash.  The concern with running under systemd is that strings that look like shell variables will not be interpolated by systemd when parsing a sysconf or `EnvironmentFile` in systemd-speak, which was being relied upon by the jenkins debian packaging.

PR #870, opened 2018-06-04
fix native type idempotency with github-oauth plugin

jhoblitt:


jhoblitt:
I'm waiting on merge to see if the acceptance tests fail [in an unusual way].  I have tested with a local jenkins instance and it has fixed the irritating non-idempotence.

jhoblitt:
travis passed... must be working ;)

PR #871, opened 2018-06-04
notify service containing class instead of service resource

jhoblitt:


jhoblitt:
This is purely a cleanup that I found in a local tree that hadn't been upstreamed.

PR #875, opened 2018-06-22
partial modulsync 1.9.2

bastelfreak:


spec/classes/cli/config_spec.rb
```diff
@@ -1,159 +1,189 @@
...
-    context 'absolute path' do
-      let(:params) { { param => '/dne' } }
-
-      it { is_expected.not_to raise_error }
-    end
-  end # validate_absolute_path
+  on_supported_os.each do |os, facts|
+    context "on #{os} " do
+      systemd_fact = case facts[:os]['family']
+                     when 'Archlinux', 'Fedora'
```
ekohl: IIRC Fedora isn't an OS family, it's part of the RedHat family.

PR #879, opened 2018-07-14
modulesync 2.4.0 + drop Ubuntu 14.04

bastelfreak:
modulesync 1.9.3

PR #882, opened 2018-07-30
Refactor repository handling

ekohl:
See the individual commits for details:

* Simplify repository handling for distros
* Use contain over anchors
* Rely on puppetlabs-apt $notify_update
* Rely on puppetlabs-apt for apt-transport-https
* Use assert_private() from stdlib
* Do not include the stdlib class

### DIFF
--- a/manifests/repo/el.pp
+++ b/manifests/repo/el.pp
@@ -1,30 +1,20 @@
 # Class: jenkins::repo::el
 #
-class jenkins::repo::el
-{
+class jenkins::repo::el {
   assert_private()
 
-  $repo_proxy = $::jenkins::repo_proxy
-
-  if $::jenkins::lts  {
-    yumrepo {'jenkins':
-      descr    => 'Jenkins',
-      baseurl  => 'https://pkg.jenkins.io/redhat-stable/',
-      gpgcheck => 1,
-      gpgkey   => 'https://pkg.jenkins.io/redhat/jenkins-ci.org.key',
-      enabled  => 1,
-      proxy    => $repo_proxy,
-    }
+  if $jenkins::lts {
+    $baseurl = "${jenkins::repo::base_url}/redhat-stable/"
+  } else {
+    $baseurl = "${jenkins::repo::base_url}/redhat/"
   }
 
-  else {
-    yumrepo {'jenkins':
-      descr    => 'Jenkins',
-      baseurl  => 'https://pkg.jenkins.io/redhat/',
-      gpgcheck => 1,
-      gpgkey   => 'https://pkg.jenkins.io/redhat/jenkins-ci.org.key',
-      enabled  => 1,
-      proxy    => $repo_proxy,
-    }
+  yumrepo {'jenkins':
+    descr    => 'Jenkins',
+    baseurl  => $baseurl,
+    gpgcheck => 1,
+    gpgkey   => "${baseurl}${jenkins::repo::gpg_key_filename}",
+    enabled  => 1,
+    proxy    => $jenkins::repo_proxy,
   }
 }
--- a/manifests/repo.pp
+++ b/manifests/repo.pp
@@ -1,7 +1,10 @@
 #
 # jenkins::repo handles pulling in the platform specific repo classes
 #
-class jenkins::repo {
+class jenkins::repo(
+  Stdlib::Httpurl $base_url = 'https://pkg.jenkins.io',
+  String $gpg_key_filename = 'jenkins.io.key',
+) {
   assert_private()
 
   if $::jenkins::repo {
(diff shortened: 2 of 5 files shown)


LongLiveCHIEF:
I pioneered a new technique for repository handling in `puppet-gitlab`, https://github.com/voxpupuli/puppet-gitlab/pull/246

Basically, it revolves around having a class param for `repository_configuration` that takes a hash, and then uses the following iterator to allow all this data to be loaded from hiera instead of a ton of code in a bunch of different puppet manifests:

```puppet
$_repository_configuration.each() | String $resource_type, Hash $resources | {
  create_resources($resource_type, $resources, $resource_defaults)
}
```

I'm not saying we shouldn't approve this PR, but just putting the pattern out there as long as we're looking at refactoring package management in this module.

Aside from drastically reducing the amount of puppet code, and simplifying repo configuration to a set of yaml files, it provides a bunch of other benefits as well:

- we can support any/all package resources without having to explicitly code them as class parameters
- we can allow our customers to be compatible with a very large range of non-direct dependency modules (such as `apt`)
- are module will require less maintenance over time as package repository resources and operating system repository systems change
- we can implicitly support both proxying _and_ mirroring together without a huge complicated logic flow for every single OS. (see https://github.com/voxpupuli/puppet-gitlab#package--repository-configuration)

jhoblitt:
In principal, I'm in favor of standardizing package repo management across voxpupuli modules as this is often a pain-point but it would be nice to have some general consensus on the method before effort is invested in a bunch of PRs.

ekohl:
I only later during refactoring realized that Jenkins repos have standardized a lot. If you look at the class params for the private repo classes they are very similar and that wasn't possible with the previous classes (note the GPG key handling). We can probably expose some of these parameters (base repo and gpg filename) but I wasn't sure we wanted to do that. Exposing parameters to private parameters was safer.

ekohl:
@LongLiveCHIEF the hard problem about that is including additional classes. For suse and RH it's no problem but for apt it's harder. Or were you proposing to include the private class here? Another issue that I'm not sure about is solving the `contain` problem for proper resource chaining.

jhoblitt:
The safest thing for the moment is probably to make the class params regular variables.

ekohl:
I've added a commit that moves parameters to the top level. I've left the Debian GPG key id as a parameter because it allows users to override it if needed. IMHO this is better because it gives users freedom. Hardcoded strings inside a classes should be avoided and a local variable that's then used once is a hardcoded string.

jhoblitt:
I still dislike that there are public api changes which, IMHO, don't feel very coherent.  This module continues to suffer from poorly designed public interfaces which are causing unnecessary maintenance burden.

As a general principle, I don't think that a module should try to accommodate every conceivable customization a user might desire but rather should provide *reasonable defaults* that can be *disabled*.  If a custom `apt`/`yum` repo is needed, presumably the end user is capable of declaring such in the manifest.  That is much more flexible approach than the limited semantics of something like `repo_base_url`.  @LongLiveCHIEF's suggested pattern provides virtually total flexibility in the configuration of alternative repos but I'd like to see some discussion among @voxpupuli/collaborators before moving in that direction.

ekohl:
I agree they're a bit awkward at the top level. I've moved them to `jenkins::repo`. They're still class parameters because in my experience it is nice if you can override things if needed. Because they're it's still a private class there are no guarantees on the forwards compatibility of those parameters.

One example when this can be nice is when the GPG key needs to be rotated because of a security issue. Now users have the ability to do this in hiera and don't have to wait on a new release.

@LongLiveCHIEF's pattern can give flexibility but I don't see it including the `apt` class which the `apt::source` resource needs. It's also a complex solution that not everybody will immediately understand. If users want that much flexibility I'd prefer to recommend disabling the repo management and doing it yourself in a profile module.

pillarsdotnet:
Looks like #879 was merged.

ekohl:
Rebased. Also submitted https://github.com/voxpupuli/puppet-jenkins/pull/917 for the non-controversial parts of this.

bastelfreak:
@ekohl can you take a look at the failed unit tests?

manifests/repo/el.pp
```diff
@@ -1,30 +1,20 @@
...
-      enabled  => 1,
-      proxy    => $repo_proxy,
-    }
+  yumrepo {'jenkins':
+    descr    => 'Jenkins',
+    baseurl  => $baseurl,
+    gpgcheck => 1,
+    gpgkey   => "${baseurl}${jenkins::repo::gpg_key_filename}",
+    enabled  => 1,
+    proxy    => $jenkins::repo::proxy,
```
rnelson0:
```suggestion
    proxy    => $jenkins::repo_proxy,
```

ekohl:
Thanks. That was a leftover from a previous iteration.

PR #884, opened 2018-08-04
[WIP] Fix rspec - preparation for #879

tuxmea:
open issues:

- Evaluation Error: Unknown variable: 'jenkins::cli::jar'. manifests/cli_helper.pp:14:14
- Puppet::Jenkins::Facts jenkins_plugins fact on Linux with plugins should eql "ant 1.2, git 2.0.1"
  expected: "ant 1.2, git 2.0.1"
  got: ""
  needs solution in spec/unit/facter/plugins_spec.rb
- how to use .sync.yml so we get mock_with at first line in
  spec_helper.rb?

### DIFF
--- a/manifests/cli/reload.pp
+++ b/manifests/cli/reload.pp
@@ -8,6 +8,8 @@ class jenkins::cli::reload {
     fail("Use of private class ${name} by ${caller_module_name}")
   }
 
+  $jar_file = $jenkins::cli::jar
+
   # Reload all Jenkins config from disk (only when notified)
   exec { 'reload-jenkins':
     command     => "${::jenkins::cli::cmd} reload-configuration",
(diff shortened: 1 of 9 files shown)


bastelfreak:
I noticed in https://github.com/voxpupuli/puppet-jenkins/pull/879 that our current .travis.yml does nothing except for printing all rake tasks during a travis job. Until the modulesync PR is merged we need to check locally if this fix actually works.

manifests/cli/reload.pp
```diff
@@ -8,13 +8,15 @@
     fail("Use of private class ${name} by ${caller_module_name}")
   }
 
+  $jar_file = $jenkins::cli::jar
```
ekohl: Why is this separate variable needed? It's only used in 1 place so I'd prefer it to be in the `require`.

tuxmea:
Please don't merge this one. This PR was initial change and changes have been adopted into #879
Unit and Acceptance Tests are green due to the reason that there are no tests.

I am closing this one as all changes have been added to #879

PR #893, opened 2018-10-22
Adding SLES to supported platforms

msurato:

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -27,6 +27,21 @@
       ]
     },
     {
+      "operatingsystem": "SLES",
+      "operatingsystemrelease": [
+        "11",
+        "12",
+        "15"
+      ]
+    },
+    {
+      "operatingsystem": "OpenSuSE",
+      "operatingsystemrelease": [
+        "42",
+        "15"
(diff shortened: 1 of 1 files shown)


othalla:
@msurato have you tested on Suse servers?
Why not adding some acceptance test for OpenSuSE?

msurato:
I have tested this on Suse, but not OpenSuSE. While they should be fine, I
would think we would want to test first. There is an issue with zypprepo
that is affecting the installation (
https://github.com/voxpupuli/puppet-zypprepo/issues/10), but putting in a
work around with installing the certificate with the RPMKey module (
https://forge.puppet.com/stschulte/rpmkey) is not difficult.

If you want, I can test with OpenSuSE and update the pull request. Please
let me know what you think! Thanks!

robpou:
+1 for this. We are using the module on SLES12 and it works (like on Rhel/Ubuntu OS).
But e.g. we manage the repository ourselves.

bastelfreak:
@robpou thanks for the info and @msurato thanks for the PR!

PR #894, opened 2018-10-26
Replace is_array with Puppet 4 native comparision

baurmatt:

### DIFF
--- a/manifests/slave.pp
+++ b/manifests/slave.pp
@@ -159,7 +159,7 @@ class jenkins::slave (
   $quoted_ui_pass = shellquote($ui_pass)
 
   if $labels {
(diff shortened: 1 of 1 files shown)


manifests/slave.pp
```diff
@@ -159,7 +159,7 @@
   $quoted_ui_pass = shellquote($ui_pass)
 
   if $labels {
-    if is_array($labels) {
+    if $labels =~ Array {
       $_combined_labels = hiera_array('jenkins::slave::labels', $labels)
```
ekohl: I do wonder what this `hiera_array` adds here, but that can be done in a separate PR.

ekohl:
Thanks!

PR #898, opened 2019-01-10
fix unreferenced variables; add puppet5/6 support

bastelfreak:

### DIFF
--- a/manifests/cli.pp
+++ b/manifests/cli.pp
@@ -40,7 +40,6 @@ class jenkins::cli {
     path        => ['/bin', '/usr/bin'],
     cwd         => '/tmp',
     refreshonly => true,
-    require     => Class['::jenkins::service'],
   }
   # Extract latest CLI in case package is updated / downgraded
   Package[$::jenkins::package_name] ~> Exec['jenkins-cli']
--- a/manifests/job/present.pp
+++ b/manifests/job/present.pp
@@ -58,6 +58,7 @@ define jenkins::job::present(
     file { $tmp_config_path:
       content => $d,
       require => Exec['jenkins-cli'],
+      before  => Exec["jenkins create-job ${jobname}"],
     }
   }
 
@@ -81,7 +82,6 @@ define jenkins::job::present(
   exec { "jenkins create-job ${jobname}":
     command => "${cat_config} | ${create_job}",
     creates => [$config_path, "${job_dir}/builds"],
-    require => File[$tmp_config_path],
   }
 
   if $replace {
@@ -91,7 +91,6 @@ define jenkins::job::present(
       command => "${cat_config} | ${update_job}",
       onlyif  => "test -e ${config_path}",
       unless  => "${difftool} ${config_path} ${tmp_config_path}",
-      require => File[$tmp_config_path],
       notify  => Exec['reload-jenkins'],
     }
   }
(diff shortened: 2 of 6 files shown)


bastelfreak:
This is WIP and needs to be rebased.

manifests/cli.pp
```diff
@@ -22,6 +22,9 @@
     Class['jenkins::service']
       -> Class['jenkins::cli']
         -> Anchor['jenkins::end']
+    $require = Class['::jenkins::service']
```
ekohl: Isn't this redundant from the 2 lines above? I think you can just leave `$require` out entirely.

manifests/job/present.pp
```diff
@@ -59,6 +60,7 @@
       content => $d,
       require => Exec['jenkins-cli'],
     }
+    $require = File[$tmp_config_path]
```
ekohl: Alternative is to add `before => Exec["jenkins create-job ${jobname}"]` on the file resource since the exec is there unconditionally

manifests/job/present.pp
```diff
@@ -91,7 +93,7 @@
       command => "${cat_config} | ${update_job}",
       onlyif  => "test -e ${config_path}",
       unless  => "${difftool} ${config_path} ${tmp_config_path}",
-      require => File[$tmp_config_path],
+      require => $require,
```
ekohl: If this requires the `create-job`, then that implicitly also requires the config file.

manifests/plugin.pp
```diff
@@ -35,6 +35,20 @@
   Any $create_user                  = undef,
 ) {
 
+  include jenkins
+
+  if $jenkins::manage_service {
+    $notify = Class['::jenkins::service']
```
ekohl: The leading colons aren't needed

manifests/sysconfig.pp
```diff
@@ -8,11 +8,16 @@
     warning("Jenkins::Sysconfig[${name}]: detected \'\$\' in value -- be advised the variable interpolation will not work under systemd")
   }
 
+  if $jenkins::manage_service {
+    $notify = Class['::jenkins::service']
+  } else {
+      $notify = undef
```
ekohl: Indenting looks a bit off

spec/defines/jenkins_plugin_spec.rb
```diff
@@ -271,6 +271,9 @@ class { jenkins:
     describe 'pin parameter' do
       context 'with pin => true' do
         let(:params) { { pin: true } }
+        let :pre_condition do
```
ekohl: Since you include it inside the plugin class, is this precondition still needed?

manifests/cli.pp
bastelfreak: true. purged it.

manifests/job/present.pp
bastelfreak: fixed it

manifests/sysconfig.pp
bastelfreak: fixed

manifests/plugin.pp
bastelfreak: fixed

spec/defines/jenkins_plugin_spec.rb
bastelfreak: fixed

mmoll:
@bastelfreak what's the status here? :)

PR #900, opened 2019-01-18
split examples about error and idempotency as much as possible

Dan33l:
This module has a large amount of acceptance code. And failing due to error on apply or due to  idempotency issue will trigger same message about failure with `works with no errors`.

The PR try to split as much as possible failing with messages:
 * `works with no errors` : when fail is only due to error.
 * `works idempotently` : when fail is only due to idempotency issue.
 * `works with no errors and idempotently` : when i was unable to split.

### DIFF
--- a/spec/acceptance/plugin_spec.rb
+++ b/spec/acceptance/plugin_spec.rb
@@ -39,20 +39,18 @@ describe 'jenkins class', order: :defined do
   end
 
   context 'default parameters' do
-    it 'works with no errors' do
-      pp = <<-EOS
-      class {'jenkins':
-        cli_remoting_free => true,
-      }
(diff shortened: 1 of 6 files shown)


spec/acceptance/plugin_spec.rb
```diff
@@ -51,7 +51,8 @@ class {'jenkins':
       }
       EOS
 
-      apply2(pp)
+      apply(pp, catch_failures: true)
```
ekohl: Why shouldn't this use apply2 as well?

Dan33l: Because ... i missed this one.
I force push a new code to fix this.

PR #904, opened 2019-03-04
Use native Puppet instead of the retries Gem in the CLI provider, replacing try_sleep parameter by exponential backoff

ekohl:
Retries is only used by the CLI providers. Puppet itself also provides retry functionality. This avoids the need to ensure a gem is installed in the right environment, which may not even work if it's a disconnected environment.

This does ignore a parameter for the maximum time to sleep. I'm looking for feedback on how to best deal with this.

### DIFF
--- a/lib/puppet/x/jenkins/provider/cli.rb
+++ b/lib/puppet/x/jenkins/provider/cli.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'puppet/provider'
+require 'puppet/util/retry_action'
 require 'facter'
 
 require 'json'
@@ -31,7 +32,6 @@ class Puppet::X::Jenkins::Provider::Cli < Puppet::Provider
   initvars
 
   commands java: 'java'
-  confine feature: :retries
 
   # subclasses should inherit this value once it has been determined that
   # jenkins requires authorization, it shortens the run time be elemating the
@@ -170,7 +170,6 @@ class Puppet::X::Jenkins::Provider::Cli < Puppet::Provider
     url                      = config[:url]
     ssh_private_key          = config[:ssh_private_key]
     cli_tries                = config[:cli_tries]
-    cli_try_sleep            = config[:cli_try_sleep]
     cli_username             = config[:cli_username]
     cli_password             = config[:cli_password]
     cli_password_file        = config[:cli_password_file]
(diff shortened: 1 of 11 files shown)


spec/unit/puppet/x/jenkins/provider/cli_spec.rb
```diff
@@ -478,15 +467,15 @@
 
     context 'network failure' do
       context 'without ssh_private_key' do
-        CLI_NET_ERRORS.each do |error|
+        CLI_AUTH_ERRORS.each do |error|
           it 'does not retry cli on AuthError exception' do
```
ekohl: In this test I went with this description of what it was supposed to do rather than the implementation.

dhollinger:
I like it, but does it create any breaking changes?

ekohl:
Maybe. From a user perspective the max total sleep time is ignored and that's a change. Perhaps we should go further and actually remove the parameter from the types. That would be a breaking change, but more honest.

lib/puppet/x/jenkins/provider/cli.rb
```diff
@@ -220,19 +219,18 @@ def self.execute_with_retry(command, options = {}, cli_pre_cmd = [])
...
+    attempt_number = 0
+    begin
       result = execute_with_auth(cli_cmd, auth_cmd, options)
       Puppet.debug("#{sname} command stdout:\n#{result}") unless result == ''
       return result
+    rescue [UnknownError, NetError] => exception
+      Puppet.debug("#{sname} caught #{exception.class.to_s.match(%r{::([^:]+)$})[1]}; retry attempt #{attempt_number}")
+      if attempt < cli_tries
+        attempt += 1
+        sleep(cli_try_sleep)
```
igalic, line 231: we could even try some exponential back-off here!

lib/puppet/x/jenkins/provider/cli.rb
```diff
@@ -27,7 +27,6 @@ def self.inherited(subclass)
   initvars
 
   commands java: 'java'
-  confine feature: :retries
```
igalic, line 30: why is that feature now gone?
does the below function not provide that feature?

igalic (commented):
🙋🏻‍♀️

lib/puppet/x/jenkins/provider/cli.rb
ekohl: Previously the feature was provided by the gem. Now it uses functionality built into Puppet itself.

igalic: you mean Ruby?
`retry` is a Ruby keyword (that i only learned about yesterday)

ekohl: Yes I did.

bastelfreak (approved):

ekohl:
It looks like the tests are failing because something doesn't pass. With the exponential backoff I think it takes too long. Probably need to tune the 30 retries a bit.

ekohl:
retries | seconds
-|-
0 | 0
1 | 0
2 | 0
3 | 0
4 | 1
5 | 2
6 | 5
7 | 12
8 | 24
9 | 50
10 | 101
11 | 203
12 | 408
13 | 817
14 | 1636
15 | 3275
16 | 6551
17 | 13105
18 | 26212
19 | 52426
20 | 104855
21 | 209713
22 | 419428
23 | 838858
24 | 1677719
25 | 3355440
26 | 6710883
27 | 13421770
28 | 26843542
29 | 53687088
30 | 107374179

I think 10 retries is a good number. 30 retries is about 3,4 years which is a bit excessive.

bastelfreak (approved):

ekohl:
It still times out where Ubuntu was previously green. I created https://github.com/voxpupuli/puppet-jenkins/pull/986 so we don't need pending anymore.

evgeni:
I've rebased this. Still need test fixes :/

bastelfreak (approved):

PR #911, opened 2019-04-26
Allow puppetlabs/apt 7.x

dhoppe:


dhoppe:
@bastelfreak I think we should merge this, even if the Travis CI pipeline is broken.

PR #912, opened 2019-05-22
Update metadata for Vox Pupuli release

pillarsdotnet:
Update urls and version numbers in the `metadata.json` file.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -1,15 +1,16 @@
 {
-  "name": "rtyler-jenkins",
+  "name": "puppet-jenkins",
   "version": "1.7.0",
   "author": "R. Tyler Croy <tyler@monkeypox.org>, Joshua Hoblitt <josh@hoblitt.com>",
   "license": "Apache-2.0",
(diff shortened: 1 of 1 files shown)


pillarsdotnet:
I don't think the test-failures are related to the metadata change.

metadata.json
```diff
@@ -59,7 +60,7 @@
     },
     {
       "name": "puppetlabs/java",
-      "version_requirement": ">= 1.0.1 < 4.0.0"
+      "version_requirement": ">= 1.0.1 < 5.0.0"
```
Dan33l: not sure failing tests are related to this. But to be sure, can you rollback to `< 4.0.0`  ?

pillarsdotnet: Done.

pillarsdotnet:
@Dan33l Tests passed.

metadata.json
pillarsdotnet: Still fails, because java 4 requires stdlib 5.  Looks like we need an update of java?

PR #915, opened 2019-06-01
release 2.0.0

bastelfreak:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,288 @@
 # Changelog
 
-This is a manually kept file, and may not entirely reflect reality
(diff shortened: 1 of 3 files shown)


Dan33l:
In the PR we are getting a indempotency error message during acceptance tests : `works idempotently (FAILED - 1)`

The second time the command that did an action was:
https://github.com/voxpupuli/puppet-jenkins/blob/master/manifests/plugin.pp#L181


This exec without condition looks guilty.

rnelson0:
Good find, @Dan33l! I agree that this looks good as is, the fix can be added later, it's really important to get 2.0.0 out there.

PR #917, opened 2019-06-06
Rely more on stdlib and apt modules

ekohl:
This reuses functionality from other modules rather than reinventing it.


bastelfreak:
Thanks. I will have a look at the failed unit test

alexjfisher:
@ekohl Thanks!

PR #919, opened 2019-07-02
Use variables for tries and try_sleep everywhere instead of hardcoding

jhooyberghs:
We sometimes run into issues where the fixed "tries 10, try_sleep: 2" on the safe-restart-jenkins isn't enough, which causes the puppetrun to fail.

We would like to be able to configure it more. There's already a cli_tries and cli_try_sleep available that fits our use case, so I reused that one.


manifests/cli.pp
```diff
@@ -66,8 +68,8 @@
   exec { 'safe-restart-jenkins':
     command     => "${cmd} safe-restart && /bin/sleep 10",
     path        => ['/bin', '/usr/bin'],
-    tries       => 10,
-    try_sleep   => 2,
```
Dan33l: The default value of `$::jenkins::cli_try_sleep` is `10` instead of `2`.
I suppose it is a simple `buxfix`

Issue #920, opened 2019-07-04
Code ordering issue on Ubuntu 18.04 and Puppet6


asifbadkarma:
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet:6
- Ruby:ruby 2.5.1p57
- Distribution:Ubuntu 18.04
- Module version:2.0.0

## How to reproduce (e.g Puppet code you use)

Called the module with the following

```
class { '::jenkins::slave':
    require                   => [ File[$jenkins_slave_dir], User[$slave_user], ],
    install_java              => false,
    masterurl                 => $jenkins_master_url,
    ui_user                   => $vault_secrets['jenkins_slave_user'],
    ui_pass                   => $vault_secrets['jenkins_slave_password'],
    slave_user                => $slave_user,
    slave_home                => $jenkins_slave_dir,
    manage_slave_user         => false,
    version                   => $swarm_plugin_version,
    slave_mode                => $slave_mode,
    executors                 => $executors,
    labels                    => $_labels,
    disable_clients_unique_id => true,
```

## What are you seeing

```
Error: Could not enable jenkins-slave: 
Error: /Service[jenkins-slave]/enable: change from 'false' to 'true' failed: Could not enable jenkins-slave: 
Error: enable: change from 'false' to 'true' failed: Could not enable jenkins-slave: 
Error: /Stage[main]/Jenkins::Slave/Jenkins::Systemd[jenkins-slave]/Transition[stop jenkins-slave service]/enable: change from enabled to transition Service[jenkins-slave] to state {"ensure"=>"stopped"} failed: enable: change from 'false' to 'true' failed: Could not enable jenkins-slave:  (corrective)
Notice: /Stage[main]/Jenkins::Slave/Service[jenkins-slave]: Dependency Transition[stop jenkins-slave service] has failures: true
Warning: /Stage[main]/Jenkins::Slave/Service[jenkins-slave]: Skipping because of failed dependencies
Warning: /Stage[main]/Jenkins::Slave/Jenkins::Systemd[jenkins-slave]/File[/etc/init.d/jenkins-slave]: Skipping because of failed dependencies
Warning: /Stage[main]/Jenkins::Slave/Jenkins::Systemd[jenkins-slave]/Systemd::Unit_file[jenkins-slave.service]/File[/etc/systemd/system/jenkins-slave.service]: Skipping because of failed dependencies
[...]

PR #921, opened 2019-07-04
Fix code ordering for ubuntu 18.04 and puppet6

asifbadkarma:

Fixes : #920

### DIFF
--- a/manifests/systemd.pp
+++ b/manifests/systemd.pp
@@ -18,10 +18,13 @@ define jenkins::systemd(
     content => template("${module_name}/${service}-run.erb"),
     owner   => $user,
     mode    => '0700',
-    notify  => Service[$service],
   }
 
-  transition { "stop ${service} service":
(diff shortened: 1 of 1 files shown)


gittygoo:
Can this be merged? seems to work for me

ekohl:
This doesn't look correct to me. My impression is that you first need to stop the old sysvinit service and then you can define the new service. Otherwise systemd will just stop the systemd based service but leave sysvinit around.

asifbadkarma:
Anything I need to do to improve this and maybe get it merged ?

ekohl:
I think your solution is incorrect. Once the systemd unit file is in place, systemd will use that to stop the service. However, the entire goal of the transition is to stop the service using the sysvinit script before removing that script. Only then can you start to set up systemd based services. While your change may appear to work, I don't think the transition has any effect anymore.

It also breaks the tests.

ekohl:
Closing for inactivity.

PR #922, opened 2019-07-08
Workaround for windows fact errors; empty plugins fact returns nil

esalberg:
Fix Windows fact error:
ERROR puppetlabs.facter - error while resolving custom facts in C:/ProgramData\PuppetLabs\puppet\cache\lib\facter\jenkins.rb: cannot load such file -- puppet/jenkins/facts

Remove empty fact jenkins_plugins for nodes not running Jenkins server

### DIFF
--- a/lib/facter/jenkins.rb
+++ b/lib/facter/jenkins.rb
@@ -4,7 +4,14 @@
 # jenkins plugins + versions.
 #
 #
+#require 'puppet/jenkins/facts'
+require 'facter'
+
+if Facter.value(:kernel) == 'windows'
+require File.join(File.dirname(__FILE__), '..', 'puppet/jenkins/facts.rb')
+else
 require 'puppet/jenkins/facts'
+end
 
 # If we're being loaded inside the module, we'll need to go ahead and add our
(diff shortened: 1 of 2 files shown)


lib/puppet/jenkins/facts.rb
```diff
@@ -33,7 +33,7 @@ def self.plugins_str
           manifest = plugins[plugin]
           buffer << "#{plugin} #{manifest[:plugin_version]}"
         end
-        buffer.join(', ')
+        buffer.join(', ') unless buffer.length == 0
```
ekohl:
```suggestion
        buffer.join(', ') unless buffer.empty?
```

lib/facter/jenkins.rb
```diff
@@ -4,7 +4,14 @@
 # jenkins plugins + versions.
 #
 #
+#require 'puppet/jenkins/facts'
```
ekohl:
```suggestion
```

ekohl:
I believe https://github.com/voxpupuli/puppet-jenkins/pull/947 should fix this in a cleaner way.

PR #929, opened 2019-08-29
Fix issue with detached plugins taking priority.

wiccan2:
As of Jenkins 2.176.2 detached plugins were taking precedence over specified plugins due to a fix that was introduced for docker: https://issues.jenkins-ci.org/browse/JENKINS-57528.

The fix in Jenkins ensured stricter processing of plugins. the module was installing plugins as `.hpi` files where Jenkins expects them to be `.jpi` files. 

This problem was highlighted in https://github.com/voxpupuli/puppet-jenkins/issues/926.

### DIFF
--- a/manifests/plugin.pp
+++ b/manifests/plugin.pp
@@ -99,7 +99,7 @@ define jenkins::plugin(
   }
 
   $plugin_ext = regsubst($download_url, '^.*\.(hpi|jpi)$', '\1')
-  $plugin     = "${name}.${plugin_ext}"
+  $plugin     = "${name}.jpi"
   # sanity check extension
   if ! $plugin_ext {
     fail("unsupported plugin extension in source url: ${download_url}")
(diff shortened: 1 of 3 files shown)


craigwatson:
Can confirm this is affecting us with 2.176.2 and 2.176.3 :)

tsde:
@wiccan2 Thanks for the PR. We're also facing the same issue. Hope this change will be merged soon. Any thought about making this change backward compatible ? We're using this module for dozens of customers and they don't necessarily follow jenkins upgrades (core or plugins) thoroughly. This change could then lead to deployment of jpi versions of plugins and I don't really know if this could break stuff.

intrigeri:
FWIW, we've faced exactly this issue at [Tails](https://tails.boum.org/) and this PR fixed it for us. Thanks!

manifests/plugin.pp
```diff
@@ -99,7 +99,7 @@
   }
 
   $plugin_ext = regsubst($download_url, '^.*\.(hpi|jpi)$', '\1')
-  $plugin     = "${name}.${plugin_ext}"
+  $plugin     = "${name}.jpi"
```
ekohl: Wouldn't it be better to download the `jpi` by default on line 97?

manifests/plugin.pp
```diff
@@ -122,29 +122,6 @@
...
-    # (up|down)grade a plugin from .jpi -> .hpi via puppet across extension
-    # changes.  Regardless, we should be relying on jenkins to guess which
-    # plugin archive to use and cleanup any conflicting extensions.
-    $inverse_plugin_ext = $plugin_ext ? {
-      'hpi'   => 'jpi',
-      'jpi'   => 'hpi',
-    }
-    $inverse_plugin     = "${name}.${inverse_plugin_ext}"
-
-    file {[
```
ekohl: Is the code to make sure it's absent still relevant?

ekohl:
I've submitted https://github.com/voxpupuli/puppet-jenkins/pull/957 as an alternative implementation which I believe is simpler.

manifests/plugin.pp
ekohl: Looks like the update URL still has hpi and this is indeed the correct thing to do.

ekohl:
This now needs a rebase. I've made some changes (https://github.com/voxpupuli/puppet-jenkins/commit/5fb0dee289343b9e32e5bcadd7c3ab11d9c751d7 and https://github.com/voxpupuli/puppet-jenkins/commit/db25b59fd2d4f5bd3b52998e781f72ff655d16c6) that at least allow you to work around this.

ekohl:
Closing for inactivity.

Issue #939, opened 2020-01-10
jenkins-slave-run script fails when JAVA_ARGS is set with multiple settings


witjoh:
This is due to the quotes used in the script when adding $JAVA_ARGS to the $SLAVE_ARGS

`/usr/bin/java '-Xms3078m -Xmx3078m -Dcom.sun.management.jmxremote -Dcom.sun.management.jmxremote.port=3000 -Dcom.sun.management.jmxremote.ssl=false -Dcom.sun.management.jmxremote.authenticate=false' -jar /opt/jenkins-slave/swarm-client-3.17.jar `
resulting in following error **Invalid initial heap size: -Xms3078m -Xmx3078m**

Removing the quotes solves the problem.  PR underway

ekohl:
This was resolved in #939

PR #940, opened 2020-01-10
Removes quotes from $JAVA_ARGS, otherwise java fails to start

witjoh:
Solves #939


vStone:
It doesn't look like the failing tests have anything to do with this change but rather the plugin system thats a bit unreliable in the tests. Could we get this one through?

ekohl:
#1013 was merged and is the same fix.

Issue #941, opened 2020-01-27
Installed Credentials + Structs plugins are not pinned to a specific version


MattWatson-:
## Problem

As stated in the title, the credentials and structs plugins are not pinned to specific versions and this is dangerous since they are installed by default

## Why
Many other plugins **_are_** pinned to a version and rely on these default plugins. So if the Jenkins server goes down and needs to be recreated, the default plugins might be installed to newer versions. The latest versions might be incompatible with the pinned plugins causing all sorts of problems. Unfortunately this happened to me recently.....

## How to solve

I think that it's fine to install the latest by default but there should be the option to override it via the init.pp but as a hash. In config.pp you should also use jenkins::plugins instead of jenkins::plugin

witjoh:
@MattWatson- check https://github.com/witjoh/puppet-jenkins_plugin if you want to manage jenkins plugins wth puppet.

PR #943, opened 2020-01-27
[#941] Allow versions of default plugins to be set

MattWatson-:

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -331,7 +331,7 @@ class jenkins(
   String $user                                    = $jenkins::params::user,
   Boolean $manage_group                           = $jenkins::params::manage_group,
   String $group                                   = $jenkins::params::group,
-  Array $default_plugins                          = $jenkins::params::default_plugins,
+  Hash $default_plugins                           = $jenkins::params::default_plugins,
   String $default_plugins_host                    = $jenkins::params::default_plugins_host,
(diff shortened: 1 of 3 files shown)


manifests/init.pp
```diff
@@ -331,7 +331,7 @@
   String $user                                    = $jenkins::params::user,
   Boolean $manage_group                           = $jenkins::params::manage_group,
   String $group                                   = $jenkins::params::group,
-  Array $default_plugins                          = $jenkins::params::default_plugins,
```
ekohl: This makes it a breaking change. Are you sure we want that?

MattWatson-:
#941

manifests/init.pp
MattWatson-: How would you explicitly set the version for the plugins in your current setup? It seems like with an array you only set the plugins but with the limitation that that you only get the latest

ekohl:
```puppet
$my_plugins = {
  'credentials' => { 'version' => 'x.y.z' },
  'structs'     =>  { 'version' => 'a.b.c' },
}
class { 'jenkins': 
  default_plugins => [],
  plugin_hash     => $my_plugins,
}
```

MattWatson-: Ah fair enough. That was an oversight on my part and I'll do that instead! If you could put that in the documentation that would be very useful. Will close the issue + PR

PR #946, opened 2020-02-21
Resolve credential leak via ps while jenkins-cli is used during puppe…

avbm:
…t runs

Passes credentials to jenkins-cli as environement variables by default so passwords aren't visible in the output of ps during puppet agent runs.

Not sure if this has an associated issue


manifests/cli.pp
```diff
@@ -64,6 +64,15 @@
     ' '
   )
 
+  if $::jenkins::cli_env_var {
+    $cmd_environment = [
+          "JENKINS_USER_ID=${::jenkins::cli_username}",
```
ekohl: Please use consistent indenting and remove the leading colons from variables. This is a repeated thing in the PR so I'll only comment it once :)
```suggestion
      "JENKINS_USER_ID=${jenkins::cli_username}",
```

manifests/init.pp
```diff
@@ -336,6 +336,48 @@
   Boolean $purge_plugins                          = $jenkins::params::purge_plugins,
 ) inherits jenkins::params {
 
+  validate_string($version)
```
ekohl: We use data types and since you're not changing anything, I think it should already be covered. Please have a look at the parameters to verify.

manifests/init.pp
```diff
@@ -401,7 +443,9 @@
     # Username / Password auth (needed for AD and other Auth Realms)
     if $_use_new_cli {
       if !empty($cli_password) {
-        $_cli_auth_arg = "-auth '${cli_username}:${cli_password}'"
+        # username and password passed as environment variables to prevent showing in ps output
+        # so setting cli_auth_arg to empty string
+        $_cli_auth_arg = ""
```
ekohl:
```suggestion
        $_cli_auth_arg = ''
```

avbm:
I made updates based on your feedback @ekohl. Let me know if it looks good now. Also I have a question about commit/merge practices. I usually like to rebase and squash updates on feature/bugfix branches. ex. my current fixes were squashed into my previous commit and pushed to my fork. I was wondering if there was a best pracitces/recommend process around this for this repo/org. (ie should we squash in fixes or keep them as separate commits)

manifests/cli.pp
avbm: fixed the indents using puppet-lint

manifests/init.pp
avbm: You are correct. I removed all the validate statements since using types makes them moot

avbm:
Rebased on current master HEAD

avbm:
@ekohl or @bastelfreak : are any other chages/updates required in this PR?
I agree with @ekohl that we need to centralize logic in a wrapper but i would prefer to do that in a separate PR to keep the changes easy to comprehend.

avbm:
@ekohl , I added a couple of unit tests to ensure the auth args are added to enviornment when defined and they are absent when not defined.
Let me know if any other tests are required else please merge.

avbm:
Hey @ekohl , just wanted to check if there were any updates required for this PR could be merged?

PR #947, opened 2020-02-22
Simplify the fact definition

ekohl:
This removes the entire module for the facts. The benefit is that it's much easier to see what's going on. There is still testing which means there's no reduction in code coverage.


lib/facter/jenkins.rb
```diff
@@ -2,10 +2,13 @@
...
+require_relative '../puppet/jenkins/plugins'
 
-# If we're being loaded inside the module, we'll need to go ahead and add our
-# facts then won't we?
-Puppet::Jenkins::Facts.install
+Facter.add(:jenkins_plugins) do
+  confine kernel: 'Linux'
+  setcode do
+    plugins = Puppet::Jenkins::Plugins.available
+    plugins.keys.sort.map { |plugin| "#{plugin} #{plugins[plugin][:plugin_version]}" }.join(', ')
```
bastelfreak: it would be so nice and easy to provide a structured fact :(

ekohl: True and it would simplify other things. IMHO we should have a `jenkins` fact that's roughly: `{jenkins: {plugins: {name1: version1, name2: version2}}}`. However, let's keep that for a later PR.

bastelfreak: I totally agree. I'm fine with merging this as is

ekohl: I started this before the big modulesync PR update. I'd prefer to rebase on that once it's merged.

PR #948, opened 2020-02-24
Remove redundant code

ekohl:
532242964375efc1013ba186f201ecbd74de0b36 changed the default from undef to false which means this code could never be hit again. Because the actual check was incorrect the tests never showed this.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -453,12 +453,8 @@ class jenkins(
     }
   }
 
-  if defined('::firewall') {
-    if $configure_firewall == undef {
-      fail('The firewall module is included in your manifests, please configure $configure_firewall in the jenkins module')
-    } elsif $configure_firewall {
-      include jenkins::firewall
-    }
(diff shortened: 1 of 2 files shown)


manifests/init.pp
```diff
@@ -453,12 +453,8 @@
     }
   }
 
-  if defined('::firewall') {
-    if $configure_firewall == undef {
-      fail('The firewall module is included in your manifests, please configure $configure_firewall in the jenkins module')
-    } elsif $configure_firewall {
-      include jenkins::firewall
-    }
+  if defined('::firewall') and $configure_firewall {
```
bastelfreak: does this refer to a fact? What about `facts('firewall')` function?

ekohl: I actually don't know for sure, but I think whether https://github.com/puppetlabs/puppetlabs-firewall/blob/master/manifests/init.pp is included (before compiling this class).

bastelfreak: okay. In that case lets keep it like it is

PR #951, opened 2020-02-24
Use more expressive rspec syntax

ekohl:


spec/classes/cli/config_spec.rb
```diff
@@ -33,43 +33,43 @@
         context 'absolute path' do
           let(:params) { { param => '/dne' } }
 
-          it { is_expected.not_to raise_error }
+          it { is_expected.to compile }
```
bastelfreak: would `compile.with_all_deps` make sense here?

ekohl: Not really IMHO. It forces a slower compilation and this test is only verifying whether the parameter is accepted.

bastelfreak: alrighty

PR #957, opened 2020-02-25
Prefer JPI plugins over HPI

### LINKED ISSUE #926: Provided plugins overwritten by detached-plugins on startup
Hi. How can the following behaviour be suppressed:

I configure a hash of plugins to be installed like so:

`
class { 'jenkins':
  plugin_hash => {
  <...>
  }
`

I would like the latest versions to be installed. But all plugins listed in the hash that are also packaged in the WAR as detached plugins get downgraded to this respective version. First, puppet correctly copies them as "hpi", but on service start they are deleted and replaced by the "jpi"s from the original WAR, virtually being downgraded.

- Puppet: 4.10.12
- Distribution: CentOS 7
- Module version: 2.0.0

ekohl:
As of Jenkins 2.176.2 detached plugins are taking precedence over specified plugins due to a fix that was introduced for docker: https://issues.jenkins-ci.org/browse/JENKINS-57528.

The fix in Jenkins ensured stricter processing of plugins. The module was installing plugins as .hpi files where Jenkins expects them to be .jpi files.

It is based on the implementation in #929 which was aimed at addressing #926.

Fixes #926
Fixes #929

### DIFF
--- a/manifests/plugin.pp
+++ b/manifests/plugin.pp
@@ -94,7 +94,7 @@ define jenkins::plugin(
 
   # if $source is specified, it overrides any other URL construction
   $download_url = $source ? {
-    undef   => "${base_url}${name}.hpi",
+    undef   => "${base_url}${name}.jpi",
     default => $source,
   }
 
--- a/spec/acceptance/plugin_spec.rb
+++ b/spec/acceptance/plugin_spec.rb
@@ -19,7 +19,7 @@ describe 'jenkins class', order: :defined do
   ].freeze
 
   shared_examples 'has_git_plugin' do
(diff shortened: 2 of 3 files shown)


ekohl:
Looks like the download URLs still have hpi.
```
  Error: Execution of '/opt/puppetlabs/puppet/bin/curl https://updates.jenkins.io/download/plugins/git/2.3.4/git.jpi -o /tmp/git.jpi_20200225-8045-1uxqkog -fsSLg --max-redirs 5' returned 22: curl: (22) The requested URL returned error: 404 Not Found
```

ekohl:
This approach isn't going to work.

PR #958, opened 2020-02-27
Fix broken acceptance tests

ekohl:
This doesn't fix all tests, but at least brings us a bit closer to working builds.


ekohl:
Still fails, but these are at least improvements to make Puppet catalogs compile so merging.

Issue #839, opened 2018-02-07
Acceptance tests related to plugin installations (or that require a plugin installation) are not failing/succeeding consistenly.


vStone:
The current acceptance tests throw a couple of errors on the installation of plugins (besides the pending tests) but do not fail the build job. Looks like the missing dependencies on the first run might get installed in further tests.

ekohl:
I got them green on Ubuntu (also in CI). For CentOS they work for me on a VM with vagrant_libvirt but in CI they fail. Not sure why. Might be docker related.

PR #962, opened 2020-04-04
modulesync 4.0.0

bastelfreak:
modulesync 2.12.0


.rubocop.yml
```diff
@@ -544,9 +544,3 @@ Style/FormatStringToken:
 # are located.
 RSpec/FilePath:
   Enabled: false
-
-RSpec/MessageSpies:
```
ekohl, line 548: For now this should be kept until someone fixes the actual code

spec/spec_helper.rb
```diff
@@ -1,85 +1,19 @@
...
-require 'rspec-puppet-facts'
-require 'bundler'
-include RspecPuppetFacts
-
-if ENV['DEBUG']
-  Puppet::Util::Log.level = :debug
-  Puppet::Util::Log.newdestination(:console)
-end
 
-# Rough conversion of grepping in the puppet source:
```
ekohl, line 18: These custom facts should be kept in. I started to work on support for this in voxpupuli-test but until then, please keep it.

ekohl (changes requested):
Needs a rebase (and probably some squashing of commits).

ekohl:
18.04 suffers from https://github.com/jenkinsci/packaging/pull/198. There's a workaround mentioned.

genebean:
looks like the plugin ownership is an issue here too @ekohl. Do you want me to dig into that?

ekohl:
> looks like the plugin ownership is an issue here too @ekohl. Do you want me to dig into that?

Please do. For me it did work in a libvirt VM on CentOS 7 and couldn't quite figure out why it fails in Docker.

I haven't looked at Ubuntu 18.04 since this PR adds it.

genebean:
Documented in #1002 and working on a fix in #1001

genebean:
@ekohl I see that the plugin stuff not being idempotent is not new (#839) - is this something we have to get fixed before moving forward or can it be a known issue for the moment? I ask because one of the reasons we started poking at this was that the currently released version does not have the right gpg keys for CentOS based servers while the code on master does. I am all for getting this fixed but it is starting to look like a non-trivial thing to deal with.

ekohl:
This looks like the same problems as we have in master so I'm merging it.

PR #967, opened 2020-05-27
Jenkins cli

anesterova:
The latest jenkins rpm package has a jenkins.war file which doesnt have jenkins-cli.jar file, that's why cli-$version.jar should be used instead. 
cli.jar file presents in older versions as well

### DIFF
--- a/lib/facter/jenkins.rb
+++ b/lib/facter/jenkins.rb
@@ -12,3 +12,8 @@ Facter.add(:jenkins_plugins) do
     plugins.keys.sort.map { |plugin| "#{plugin} #{plugins[plugin][:plugin_version]}" }.join(', ')
   end
 end
+Facter.add(:jenkins_version) do
+    setcode do
+        Facter::Util::Resolution.exec("rpm -qa|grep jenkins|cut -d'-' -f2")
+    end
+end
--- a/lib/facter/jenkins.rb
+++ b/lib/facter/jenkins.rb
@@ -12,8 +12,3 @@ Facter.add(:jenkins_plugins) do
     plugins.keys.sort.map { |plugin| "#{plugin} #{plugins[plugin][:plugin_version]}" }.join(', ')
   end
 end
-Facter.add(:jenkins_version) do
-    setcode do
-        Facter::Util::Resolution.exec("rpm -qa|grep jenkins|cut -d'-' -f2")
-    end
-end
--- a/manifests/cli.pp
+++ b/manifests/cli.pp
@@ -25,7 +25,6 @@ class jenkins::cli {
 
   $jar = "${jenkins::libdir}/cli.jar"
   $extract_jar = "jar -xf ${jenkins::libdir}/jenkins.war WEB-INF/lib/"
-  #$extract_jar = "jar -xf ${jenkins::libdir}/jenkins.war WEB-INF/lib/cli-${jenkins_version}.jar"
   $move_jar = "mv WEB-INF/lib/cli-*.jar ${jar}"
   $remove_dir = 'rm -rf WEB-INF'
   $cli_tries = $jenkins::cli_tries
--- a/manifests/cli.pp
+++ b/manifests/cli.pp
@@ -23,9 +23,10 @@ class jenkins::cli {
   }
 
 
-  $jar = "${jenkins::libdir}/cli-${::jenkins_version}.jar"
-  $extract_jar = "jar -xf ${jenkins::libdir}/jenkins.war WEB-INF/lib/cli-${jenkins_version}.jar"
-  $move_jar = "mv WEB-INF/lib/cli-${::jenkins_version}.jar ${jar}"
+  $jar = "${jenkins::libdir}/cli.jar"
+  $extract_jar = "jar -xf ${jenkins::libdir}/jenkins.war WEB-INF/lib/"
+  #$extract_jar = "jar -xf ${jenkins::libdir}/jenkins.war WEB-INF/lib/cli-${jenkins_version}.jar"
+  $move_jar = "mv WEB-INF/lib/cli-*.jar ${jar}"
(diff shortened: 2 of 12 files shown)


lib/facter/jenkins.rb
```diff
@@ -12,3 +12,8 @@
     plugins.keys.sort.map { |plugin| "#{plugin} #{plugins[plugin][:plugin_version]}" }.join(', ')
   end
 end
+Facter.add(:jenkins_version) do
+    setcode do
+        Facter::Util::Resolution.exec("rpm -qa|grep jenkins|cut -d'-' -f2")
```
ekohl, line 17: This only works on Red Hat based systems. It'll also fail on the first run because the RPM is not yet installed. On upgrades it will fail because of a mismatch

manifests/cli.pp
```diff
@@ -22,9 +22,10 @@
         -> Anchor['jenkins::end']
   }
 
-  $jar = "${jenkins::libdir}/jenkins-cli.jar"
-  $extract_jar = "jar -xf ${jenkins::libdir}/jenkins.war WEB-INF/jenkins-cli.jar"
-  $move_jar = "mv WEB-INF/jenkins-cli.jar ${jar}"
+
+  $jar = "${jenkins::libdir}/cli-${::jenkins_version}.jar"
+  $extract_jar = "jar -xf ${jenkins::libdir}/jenkins.war WEB-INF/lib/cli-${jenkins_version}.jar"
```
ekohl, line 27: Wouldn't it be better to list the files in the war and grep the exact file name rather than relying on a fact?

lib/facter/jenkins.rb
anesterova: yes, sorry, you are right

manifests/cli.pp
anesterova: we can unzip and grep yes, will do

anesterova: @ekohl check the latest commit, that should work.
```
root@lga-cmpjenkins01 anesterova # jar -xf /usr/lib/jenkins/jenkins.war WEB-INF/lib/

root@lga-cmpjenkins01 anesterova # ll

drwxr-xr-x 3 root root   16 May 27 14:52 WEB-INF

root@lga-cmpjenkins01 anesterova # mv WEB-INF/lib/cli-*.jar /tmp/cli.jar

root@lga-cmpjenkins01 anesterova # ll /tmp/cli.jar

-rw-r--r-- 1 root root 3139801 May 18 14:50 /tmp/cli.jar

root@lga-cmpjenkins01 anesterova # ```

**Comment by ekohl** [comment]
This approach looks good, let's see if the acceptance tests agree. The unit tests need to be updated to reflect these changes, but that's to be expected.

File: manifests/cli.pp [hunk 3]
```diff
@@ -22,9 +22,10 @@
         -> Anchor['jenkins::end']
   }
 
-  $jar = "${jenkins::libdir}/jenkins-cli.jar"
-  $extract_jar = "jar -xf ${jenkins::libdir}/jenkins.war WEB-INF/jenkins-cli.jar"
-  $move_jar = "mv WEB-INF/jenkins-cli.jar ${jar}"
+
+  $jar = "${jenkins::libdir}/cli.jar"
```
ekohl, line 26: What's the reason this needs to change? The overall change would be smaller if we didn't and since we control this file, there should be no issue. Am I missing something?

anesterova: do you mean why the name is cli.jar instead of jenkins-cli.jar?

ekohl: Yes

anesterova: > Do you happen to know which version introduced this change? It'd be nice to set a minimum version number in the README.
not sure, the latest i had was 2.210-1.1 and it worked with jenkins-cli.jar

manifests/cli.pp
```diff
@@ -22,9 +22,10 @@
...
+  $extract_jar = "jar -xf ${jenkins::libdir}/jenkins.war WEB-INF/lib/"
+  $move_jar = "mv WEB-INF/lib/cli-*.jar ${jar}"
```
ekohl, line 28: Thinking out loud: would this work and provide more compatibility or only ask for more issues
```suggestion
  $move_jar = "mv WEB-INF/lib/*cli*.jar ${jar}"
```

manifests/cli.pp
anesterova: > Yes

no reason. Thought the origin file was cli-$version.jar, that's why renamed. Can change it back if you want.

ekohl: I'm always a fan of the smallest possible changes. That makes it easier to review.

manifests/cli.pp
anesterova: if they decide to add some xyz-cli-xyz.jar we might have problems

manifests/cli.pp
anesterova: done

manifests/cli.pp
ekohl: That's exactly what I meant with `ask for more issues`. I'm fine with the current version, but it deserves a note in README.md

anesterova: i dont think there is a minimum version number which should be mentioned. In old jenkins packages both jenkins-cli and cli-version exist

anesterova: the oldest version i checked was jenkins-2.99-1.1

ekohl: Thanks for checking.

manifests/cli.pp
```diff
@@ -22,9 +22,10 @@
         -> Anchor['jenkins::end']
   }
 
+
```
ekohl, line 25:
```suggestion
```

ekohl:
I've taken the liberty to squash the commits and write a slightly more descriptive commit message. Thanks!

PR #968, opened 2020-05-28
Use eval in jenkins-run

stek29:
It's required to support escaping spaces in JENKINS_JAVA_CMD, JENKINS_JAVA_OPTIONS, JENKINS_HOME and JENKINS_WAR variables.

For example, it's impossible to set [`hudson.model.DirectoryBrowserSupport.CSP`](https://wiki.jenkins.io/display/JENKINS/Configuring+Content+Security+Policy) without this, since CSP policy has to contain spaces, and escaping spaces wasn't possible in `JENKINS_JAVA_OPTS`.

### DIFF
--- a/templates/jenkins-run.erb
+++ b/templates/jenkins-run.erb
@@ -75,4 +75,4 @@ if [ "$JENKINS_ENABLE_ACCESS_LOG" = "yes" ]; then
   PARAMS+=("--simpleAccessLogger.file=/var/log/jenkins/access_log")
 fi
 
-$JAVA_CMD "${PARAMS[@]}"
+eval "$JAVA_CMD ${PARAMS[@]}"


templates/jenkins-run.erb
```diff
@@ -75,4 +75,4 @@ if [ "$JENKINS_ENABLE_ACCESS_LOG" = "yes" ]; then
   PARAMS+=("--simpleAccessLogger.file=/var/log/jenkins/access_log")
 fi
 
-$JAVA_CMD "${PARAMS[@]}"
```
ekohl, line 78: It is my understanding that this way of using bash arrays actually should have the proper quotes.

stek29: it's not about quotes in PARAMS array, it's about quotes in JENKINS_JAVA_CMD , JENKINS_JAVA_OPTIONS, JENKINS_HOME or JENKINS_WAR.

stek29:
It can't be solved on line 53: if it's quoted on line 53, since it's used without quotes on line 78.

If you look at most of other scripts written in this way, you'll see that it's a common pattern to use some wrapper to start the command (like `daemon` function in init.d scripts), or eval if there's no specific wrapper.

I wasn't able to find any way to set JENKINS_JAVA_OPTS to some value which has spaces in it without using eval on line 78.

notice: JENKINS_JAVA_OPTS has to be a string, it can't be an array (and there's no way to change this, at least without dropping Systemd env file and switching to `source` in run script).
so, it has to be split on spaces. the way to  stop it from splitting when you don't want it to is to escape spaces, or have escaped quotes, and the only way to work with that is using eval + quoted string.

ekohl:
> notice: JENKINS_JAVA_OPTS has to be a string, it can't be an array (and there's no way to change this, at least without dropping Systemd env file and switching to `source` in run script).
> so, it has to be split on spaces. the way to stop it from splitting when you don't want it to is to escape spaces, or have escaped quotes, and the only way to work with that is using eval + quoted string.

Can you elaborate this? I may be missing something, but won't this work as the last line:

```bash
$JENKINS_JAVA_CMD "$JENKINS_JAVA_OPTIONS" -DJENKINS_HOME=$JENKINS_HOME -jar $JENKINS_WAR "${PARAMS[@]}"
```

Simple example of why I think this works:
```console
$ cat test.sh 
#!/bin/bash
set -x
echo "$BLA"
$ BLA='"abc" "def"' ./test.sh 
+ echo '"abc" "def"'
"abc" "def"
$ BLA="'abc' 'def'" ./test.sh 
+ echo ''\''abc'\'' '\''def'\'''
'abc' 'def'
```

stek29:
```
~
❯ cat test.sh
#!/bin/bash
printf '%s\n' "$BLA"
eval printf '%s\\n' "$BLA"

~
❯ BLA="'abc' 'def' gk\ lm" ./test.sh
'abc' 'def' gk\ lm
abc
def
gk lm
```

stek29:
@ekohl was that example sufficient?

ekohl:
I'll be mostly offline for the next 2 weeks. Some other reviewer should feel free to jump in.

ekohl:
Made obsolete in #1035

PR #980, opened 2020-08-25
Fix GPG URL key for Centos7

### LINKED ISSUE #979: Jenkins not installing on Centos 7 due to gpg key
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 5.5.21
- Ruby: 
- Distribution: Centos7
- Module version: 2.0.0

## How to reproduce (e.g Puppet code you use)
Install a new server on Centos7

## What are you seeing
Jenkins doesn't install to fix have to do the following: 
```
sudo wget -O /etc/yum.repos.d/jenkins.repo https://pkg.jenkins.io/redhat/jenkins.repo
sudo rpm --import https://pkg.jenkins.io/redhat/jenkins.io.key
```

We've tried many fresh installs and we still get the same issue. I've changed this manually in the module in our puppet code and the fix work. As the previous key in the 2.0.0 version is this:

https://pkg.jenkins.io/redhat/jenkins-ci.org.key

And it should be:

https://pkg.jenkins.io/redhat/jenkins.io.key

## What behaviour did you expect instead
Jenkins is just installed

Wade9320:

Updates problems with gpg key when installing on CentOS

Fixes #979

### DIFF
--- a/manifests/repo/el.pp
+++ b/manifests/repo/el.pp
@@ -6,14 +6,14 @@ class jenkins::repo::el {
   if $jenkins::lts {
     $baseurl = "${jenkins::repo::base_url}/redhat-stable/"
   } else {
-    $baseurl = "${jenkins::repo::base_url}/redhat/"
+    $baseurl = "${jenkins::repo::base_url}/redhat/${jenkins::repo::gpg_key_filename}"
   }
 
   yumrepo {'jenkins':
     descr    => 'Jenkins',
     baseurl  => $baseurl,
     gpgcheck => 1,
-    gpgkey   => "${baseurl}${jenkins::repo::gpg_key_filename}",
+    gpgkey   => "${baseurl}/redhat/${jenkins::repo::gpg_key_filename}",
     enabled  => 1,
     proxy    => $jenkins::repo_proxy,
   }


manifests/repo/el.pp
```diff
@@ -6,14 +6,14 @@
   if $jenkins::lts {
     $baseurl = "${jenkins::repo::base_url}/redhat-stable/"
   } else {
-    $baseurl = "${jenkins::repo::base_url}/redhat/"
+    $baseurl = "${jenkins::repo::base_url}/redhat/${jenkins::repo::gpg_key_filename}"
```
ekohl, line 9: This looks wrong

manifests/repo/el.pp
```diff
@@ -6,14 +6,14 @@
...
   }
 
   yumrepo {'jenkins':
     descr    => 'Jenkins',
     baseurl  => $baseurl,
     gpgcheck => 1,
-    gpgkey   => "${baseurl}${jenkins::repo::gpg_key_filename}",
+    gpgkey   => "${baseurl}/redhat/${jenkins::repo::gpg_key_filename}",
```
ekohl, line 16: This doesn't looks right either. You end up with `/redhat/redhat/jenkins.io.key`.

ekohl (changes requested):
I don't think this entire PR is valid. The problem you are seeing is that the GPG key as used by Jenkins has changed. https://pkg.jenkins.io/redhat/ states:
> WARNING: The gpg key use to sign our packages has been updated on 16th of April 2020, therefore you need to reimport it if you imported before this date. 

There is no migration for that in this module and yum doesn't fetch the key again unless explicitly instructed to do so.

Note this is only a problem on upgrades, fresh installs should still work because it fetches the key.

Wade9320:
We've tried many fresh installs and we still get the same issue. I've changed this manually in the module in our puppet code and the fix work. As the previous key in the 2.0.0 version is this:

https://pkg.jenkins.io/redhat/jenkins-ci.org.key

And it should be: 

https://pkg.jenkins.io/redhat/jenkins.io.key

Wade9320:
Yes you're correct @ekohl this PR isn't needed but the problem in #979 still persists in version 2.0.0

PR #981, opened 2020-08-27
Fix the plugin acceptance tests

ekohl:


bastelfreak (approved):

ekohl:
When I run it locally it does pass. Going to merge since it at least makes Ubuntu green again so that's an improvement.

PR #984, opened 2020-09-03
Drop code old CLI remoting support

ekohl:
Jenkins 2.54 was released in 2017, which is long EOL. Current versions have even dropped support for the old remoting.


bastelfreak (approved):

llowder (approved):

ekohl:
Fixed tests so they now only fail in the same way as master. Merging since there's been no code changes outside of tests since the approval.

PR #986, opened 2020-09-07
Implement ConduitCredentialsImpl in groovy_helper

ekohl:
This is entirely untested and I've never used this plugin, but I want to get rid of the pending tests.


bastelfreak (approved):

bastelfreak (approved):

ekohl:
Ok, I think it should now pass and get at least that test green.

PR #987, opened 2020-09-08
Drop GoogleRobotPrivateKeyCredentials implementation

ekohl:
This implementation is broken and as preparation for 3.0.0 the code base is being reduced to working versions. If this is desired by users, a working patch should be provided.

I tried to get it to work, but I couldn't get it to work quickly. It looks like you can no longer get the private data.


bastelfreak (approved):
always :+1: for dropping legacy stuff

igalic:
where are the failing tests coming from?

ekohl:
> always +1 for dropping legacy stuff

It's not really legacy since the plugin is still maintained. It's just that our implementation is broken and I couldn't fix it very quickly.

> where are the failing tests coming from?

Looks like I forgot to clean up some code.

spec/unit/puppet/provider/jenkins_credentials/cli_spec.rb
```diff
@@ -284,74 +254,6 @@
...
-        json_key
-      ].each do |k|
-        expect(provider.public_send(k.to_sym)).to eq :absent
-      end
-    end
-  end
-
-  shared_examples 'a provider from example hash 7' do
-    it do
-      cred = credentials[5]
```
ekohl, line 309: Fun fact: this has a nice copy paste error: hash 7 maps to entry 5 instead of 6. It never showed because these samples don't run.

ekohl:
The failures are now the same as in master.

PR #992, opened 2020-10-07
Change hiera_array to the newer lookup

### LINKED ISSUE #983: function hiera_array is deprecated
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 6.17.0
- Ruby: 2.0.0
- Distribution: CentOS Linux release 7.8.2003
- Module version: 2.0.0

## Output log
Puppet The function 'hiera_array' is deprecated in favor of using 'lookup'. See https://puppet.com/docs/puppet/6.17/deprecated_language.html

https://github.com/voxpupuli/puppet-jenkins/blob/c0eaac41e0e3218b5f832ba0980c372948bb03c8/manifests/slave.pp#L163
https://github.com/voxpupuli/puppet-jenkins/blob/c0eaac41e0e3218b5f832ba0980c372948bb03c8/manifests/slave.pp#L173
https://github.com/voxpupuli/puppet-jenkins/blob/c0eaac41e0e3218b5f832ba0980c372948bb03c8/manifests/slave.pp#L183

drzombey:
This PR remove the deprecated hiera_array and change it to the new lookup.

Fixes #983

### DIFF
--- a/manifests/slave.pp
+++ b/manifests/slave.pp
@@ -160,7 +160,7 @@ class jenkins::slave (
 
   if $labels {
     if $labels =~ Array {
-      $_combined_labels = lookup( { 'name' => 'jenkins::slave::labels', 'default_value' => $labels } )
+      $_combined_labels = lookup('jenkins::slave::labels', Any, unique, $labels)
       $_real_labels = join($_combined_labels, ' ')
     }
     else {
@@ -170,7 +170,7 @@ class jenkins::slave (
 
   if $java_args {
     if $java_args =~ Array {
-      $_combined_java_args = lookup( { 'name' => 'jenkins::slave::java_args', 'default_value' => $java_args } )
+      $_combined_java_args = lookup('jenkins::slave::java_args', Any, unique, $java_args)
       $_real_java_args = join($_combined_java_args, ' ')
     }
     else {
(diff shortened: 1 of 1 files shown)


manifests/slave.pp
```diff
@@ -160,7 +160,7 @@
 
   if $labels {
     if $labels =~ Array {
-      $_combined_labels = hiera_array('jenkins::slave::labels', $labels)
+      $_combined_labels = lookup( { 'name' => 'jenkins::slave::labels', 'default_value' => $labels } )
```
ekohl, line 163: I wonder if this is really needed. The hiera key `jenkins::slave::labels` means `$labels` already contains the correct content, right? I think that you can simply drop `hiera_array` everywhere.

alexjfisher: Not quite.  Specifically, `hiera_array` would have been merging arrays across the hierarchy.

alexjfisher: You need to either set the merge behaviour as a parameter to `lookup` here, or set lookup_options data in module hiera.  See https://github.com/voxpupuli/puppet-telegraf/pull/134 for example.

alexjfisher (changes requested):

manifests/slave.pp
drzombey: You mean something like this `lookup('jenkins::slave::labels', Any, unique, $labels)` ?

alexjfisher: yeah, something like is needed to preserve behaviour.  Bonus points if you can rid of it and use `lookup_options` in module hiera though.  That is more flexible as users can override the merge behaviour in their own environment hiera if they fancy.

But don't worry if you can't get that working.  Switching to `lookup()` would still be an improvement.

ekohl:
Replaced by https://github.com/voxpupuli/puppet-jenkins/pull/1055

PR #996, opened 2020-11-05
Always read plugin manifest as UTF-8

ekohl:
The manifest can contain UTF-8 data and if no LANG is set (like in minimal environments), this causes the read to fail. This forces it to read the file using UTF-8 encoding.


bastelfreak (approved):

ekohl:
The previous build failed, but it looks like there were 2 occurrences of `File.read`. Trying with the other one too.

ekohl:
It doesn't help to fix the EL7 tests under docker :(

PR #997, opened 2020-11-27
Feature/oic auth

zipkid:
This PR adds support for OpenId Connect Authentication https://plugins.jenkins.io/oic-auth/ securityRealm.

I am not very happy with the use of the "Boolean:<true/false>" construct that i added for boolean parameters.
This was necessary because one of the parameters for the OicSecurityRealm constructor is a classed Boolean opposed to all others being primitive booleans. I am not very fluent in Java so i did not know how to solve this a different way. If anyone can provide a more elegant way around this issue i would love to change it.

### DIFF
--- a/files/puppet_helper.groovy
+++ b/files/puppet_helper.groovy
@@ -143,8 +143,17 @@ class Util {
       }
     } else {
       // XXX explicit type declaration is required here
+      // def Class[] signature = args.collect {
+      //   it instanceof Boolean ? boolean.class : it.class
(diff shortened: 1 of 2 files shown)


zipkid:
I don't find how the travis failures on Centos are related to my changes.

ekohl:
They already were failing. I haven't been able to pinpoint why. When I run the tests on a Vagrant VM (using libvirt) they pass but on Docker containers they fail.

PR #1001, opened 2021-01-08
Set plugin ownership, remove duplicate include

genebean:
This change ensures that the proper ownership is set for the downloaded
plugin. It also removes a duplicate include of the base jenkins class.


ekohl (approved):

ekohl:
If you have commit access, then you should be able to push to the `modulesync` branch directly. That's what I usually do.

genebean:
I don't have any rights in the voxpupuli org, just over in Puppet's orgs.

genebean:
This also let's me test in a sane way w/o cluttering up that branch with mess :)

ekohl:
Fair enough. If you want, we welcome new contributors :)

manifests/plugin.pp
```diff
@@ -194,9 +192,18 @@
...
+      group           => $jenkins::group,
       require         => $plugindir,
       notify          => $notify,
     }
+
+    exec { "${plugin} permissions":
+      command   => "chown ${jenkins::user}:${jenkins::group} ${jenkins::plugin_dir}/${plugin}",
+      path      => '/usr/sbin:/usr/bin',
+      subscribe => Archive[$plugin],
+    }
```
ekohl, line 205: Can't this be done with a file resource?

You'd also think that this part takes care of it already:
https://github.com/voxpupuli/puppet-archive/blob/ac5a8a4e756499dfa315efe2a63abe6383bd00ca/lib/puppet_x/bodeco/archive.rb#L34-L57

genebean: The exec is straight out of archive's readme fwiw since we don't crate the plugins dir ourselves.

ekohl: I've used the file resource in similar ways before. Not with archive, but where native type ends up creating a directory. As long as the dependency is there, it's fine to add one and avoids an exec (which is always a good thing to avoid when possible).

genebean:
> Fair enough. If you want, we welcome new contributors :)

I am happy to become a contributor :)

ekohl:
Btw, my theory was that it's unable to read the list of installed plugins in the Docker environment and fails because of that. It appears to be trying to install structs but that's present out of the box. Maybe the fact fails in some way.

manifests/plugin.pp
genebean: so, the exec did resolve the issue I was seeing where permissions needed to be fixed. There seems to be a second layer to this though, and the file resource might not have been the root of why there was a second run needed - it seem that there might be an ordering issue as it kinda looks like `jenkins::config` is executing its ensure_resource on the credentials plugin (maybe others too) on the second run.

ekohl: There's a [fact](https://github.com/voxpupuli/puppet-jenkins/blob/master/lib/facter/jenkins.rb) that fails. Perhaps this part is broken in the container:
https://github.com/voxpupuli/puppet-jenkins/blob/15ffa7160bf92546c9246d4a7644457ed06732df/lib/puppet/jenkins.rb#L6

genebean: ooooo that could do it... looking

genebean: I may be having a memory lapse, but wouldn't the fact always fail on the first run as that user does not exist pre-manifest application when facts are evaluated?

ekohl: Oh yes, that's very likely. Perhaps it should have a fallback to `/var/lib/jenkins` which AFAIK is the default in packaging.

ekohl:
Looks like I closed this because I deleted the modulesync branch. I also opened #1004 to try out some other theories but they don't appear to be working.

PR #1003, opened 2021-01-11
Change supported Puppet versions to 6 & 7

genebean:

With Puppet 5 EoL and Puppet 7 out, I figured it was time to bump metadata.json too.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -83,7 +83,7 @@
   "requirements": [
     {
       "name": "puppet",
-      "version_requirement": ">= 5.5.8 < 7.0.0"
+      "version_requirement": ">= 6.1.0 < 8.0.0"
     }
   ]
 }


genebean:
The failure seems to be the same as what is on master here

metadata.json
```diff
@@ -83,7 +83,7 @@
   "requirements": [
     {
       "name": "puppet",
-      "version_requirement": ">= 5.5.8 < 7.0.0"
+      "version_requirement": ">= 6.0.0 < 8.0.0"
```
ekohl, line 86: https://github.com/voxpupuli/voxpupuli.github.io/issues/229 is still open, but I think the minimum version should be 6.1.0 since that doesn't require `systemctl daemon-reload` which is nice.

ekohl (approved):

PR #1004, opened 2021-01-12
Assume the package default for Jenkins homedir

ekohl:
In case the Jenkins user doesn't exist, we can assume what's default in the package because it may be present in the future.

This is a test to see if this fixes the acceptance tests on EL7 when running under docker.


ekohl:
We got the tests working again, but there is a useful part in here: https://github.com/voxpupuli/puppet-jenkins/pull/1056

PR #1007, opened 2021-01-27
download jenkins-cli.jar instead of extracting it from jenkins.war

hoomaan-kh:
… the Jenkins-cli.

https://github.com/voxpupuli/puppet-jenkins/issues/998

### DIFF
--- a/manifests/cli.pp
+++ b/manifests/cli.pp
@@ -23,11 +23,7 @@ class jenkins::cli {
   }
 
   $jar = "${jenkins::libdir}/jenkins-cli.jar"
-  $extract_jar = "jar -xf ${jenkins::libdir}/jenkins.war WEB-INF/lib/"
-  $move_jar = "mv WEB-INF/lib/cli-*.jar ${jar}"
-  $remove_dir = 'rm -rf WEB-INF'
-  $cli_tries = $jenkins::cli_tries
-  $cli_try_sleep = $jenkins::cli_try_sleep
+  $download_jar = "wget -O ${jar} http://localhost:8080/jnlpJars/jenkins-cli.jar"
 
   # make sure we always call Exec[jenlins-cli] in case
   # the binary does not exist
@@ -36,7 +32,7 @@ class jenkins::cli {
     creates => $jar,
   }
   ~> exec { 'jenkins-cli' :
-    command     => "${extract_jar} && ${move_jar} && ${remove_dir}",
+    command     => "sleep 30 && ${download_jar}",
     path        => ['/bin', '/usr/bin'],
     cwd         => '/tmp',
     refreshonly => true,


ekohl:
This has a merge conflict. Could you make sure it cleanly applies?

hoomaan-kh:
The conflict had been fixed

manifests/cli.pp
```diff
@@ -36,7 +32,7 @@
     creates => $jar,
   }
   ~> exec { 'jenkins-cli' :
-    command     => "${extract_jar} && ${move_jar} && ${remove_dir}",
+    command     => "sleep 30 && ${extract_jar}",
```
kenyon, line 35: Why `sleep 30`?

manifests/cli.pp
```diff
@@ -23,11 +23,7 @@
   }
 
   $jar = "${jenkins::libdir}/jenkins-cli.jar"
-  $extract_jar = "jar -xf ${jenkins::libdir}/jenkins.war WEB-INF/lib/"
-  $move_jar = "mv WEB-INF/lib/cli-*.jar ${jar}"
-  $remove_dir = 'rm -rf WEB-INF'
-  $cli_tries = $jenkins::cli_tries
-  $cli_try_sleep = $jenkins::cli_try_sleep
+  $extract_jar = "/bin/wget -O ${jar} http://127.0.0.1:8080/jnlpJars/jenkins-cli.jar"
```
kenyon, line 26: If we're not extracting anymore, we shouldn't call the variable `$extract_jar`.

kenyon, line 26: Should be `wget`, not `/bin/wget`.

kenyon (changes requested):
I've rebased to remove the merge commit, and fixed the commit message, but I'm not sure about the actual changes.

manifests/cli.pp
kenyon, line 26: The host part of the URL should be `localhost`, not `127.0.0.1`, so that we are not dependent on legacy IPv4 being available.

kenyon:
For reference, the docs for this are here: https://www.jenkins.io/doc/book/managing/cli/#downloading-the-client

manifests/cli.pp
hoomaan-kh: As it is downloading the "jenkins-cli" right after the Jenkins service is activated, we need to wait for a couple of seconds so the path will be accessible.

manifests/cli.pp
hoomaan-kh: I'll change the name

hoomaan-kh:
Do we need more changes or it could be merged?

kenyon:
> Do we need more changes or it could be merged?

We need the tests to pass. I rebased on latest master and squashed the commits.

ekohl:
The current solution does work and there hasn't been activity for a year. Closing, but feel free to open a new PR.

PR #1010, opened 2021-04-16
puppet/archive: allow 5.x

bastelfreak:

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -69,7 +69,7 @@
     },
     {
       "name": "puppet/archive",
-      "version_requirement": ">= 1.3.0 < 5.0.0"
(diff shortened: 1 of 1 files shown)


ekohl (commented):
Any particular reason you're moving the block instead of replacing it inline?

bastelfreak:
I was hoping that this causes the puppet module tool to resolve the dependencies, but that doesn't work :(

bastelfreak:
This depends on https://github.com/puppetlabs/puppetlabs-java/pull/476

saz:
@bastelfreak puppetlabs/puppetlabs-java#476 is merged now

PR #1025, opened 2021-12-06
add version fact and daemon param for systemd unit file

fe80:
Hello,

This PR add a new fact for the jenkins version and daemon param.

Define a systemd type param (forking or simple)

Actually no spec is add in the systemd content, I can add a unit test for that if you want.

https://github.com/voxpupuli/puppet-jenkins/issues/1024

### DIFF
--- a/lib/facter/jenkins.rb
+++ b/lib/facter/jenkins.rb
@@ -12,3 +12,25 @@ Facter.add(:jenkins_plugins) do
     plugins.keys.sort.map { |plugin| "#{plugin} #{plugins[plugin][:plugin_version]}" }.join(', ')
   end
 end
+
+Facter.add(:jenkins_version) do
+  confine kernel: 'Linux'
+
+  setcode do
+    libdir = case Facter.value('os.family')
+             when 'Debian'
+               '/usr/share/jenkins'
+             when 'Archlinux'
+               '/usr/share/java/jenkins'
+             else
+               '/usr/lib/jenkins'
+             end
+    war = libdir + '/jenkins.war'
+
+    if Facter::Util::Resolution.which('java') and File.exist?(war)
+      Facter::Util::Resolution.exec(
+        'java -jar %s --version' % [war]
+      )
+    end
+  end
+end
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -30,6 +30,12 @@ class jenkins::params {
   ]
   $purge_plugins = false
 
+  if versioncmp(pick($facts['jenkins_version'], '2.313'), '2.313') >= 0 {
+    $systemd_type = 'simple'
+  } else {
+    $systemd_type = 'forking'
+  }
+
   case $facts['os']['family'] {
     'Debian': {
       $repo                 = true
--- a/spec/classes/jenkins_service_spec.rb
+++ b/spec/classes/jenkins_service_spec.rb
@@ -51,9 +51,10 @@ describe 'jenkins' do
             end
             it do
               is_expected.to contain_systemd__unit_file('jenkins.service').
-                that_notifies('Service[jenkins]')
+                that_notifies('Service[jenkins]').with_content(/^Type=simple/)
             end
           end
+
         when '6'
           context 'EL 6' do
             it do
@@ -73,6 +74,27 @@ describe 'jenkins' do
                 enable: false
               )
             end
+          end
+        end
+      end
+      if os_facts[:os]['release']['major'].to_i >= 7
+        context 'with jenkins version 2.312' do
+          let(:facts) { os_facts.merge!({ jenkins_version: '2.312' }) }
+
+          it do
+            is_expected.to contain_systemd__unit_file('jenkins.service').
+              with_content(%r{^Type=forking})
+          end
+        end
+      end
+
+      if os_facts[:os]['release']['major'].to_i >= 7
+        context 'with jenkins version 2.313' do
+          let(:facts) { os_facts.merge!({ jenkins_version: '2.313' }) }
+
+          it do
+            is_expected.to contain_systemd__unit_file('jenkins.service').
+              with_content(%r{^Type=simple})
           end
         end
       end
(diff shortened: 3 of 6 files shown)


lib/facter/jenkins.rb
```diff
@@ -12,3 +12,24 @@
     plugins.keys.sort.map { |plugin| "#{plugin} #{plugins[plugin][:plugin_version]}" }.join(', ')
   end
 end
+
+Facter.add(:jenkins_version) do
+  confine kernel: 'Linux'
+
+  setcode do
+    war = case Facter.value('os')['family']
```
ekohl, line 20: Doesn't this also work?
```suggestion
    war = case Facter.value('os.family')
```

lib/facter/jenkins.rb
```diff
@@ -12,3 +12,24 @@
...
+          when 'Debian'
+            '/usr/share/jenkins'
+          when 'Archlinux'
+            '/usr/share/java/jenkins/'
```
ekohl, line 24: Why is there a trailing slash here?

lib/facter/jenkins.rb
```diff
@@ -12,3 +12,24 @@
...
+          else
+            '/usr/lib/jenkins'
+          end + '/jenkins.war'
```
ekohl, line 27: This bit is very well hidden. I think it would be easier to read if you have a `directory = ...` and `war = directory + "/jenkins.war"`.

manifests/params.pp
```diff
@@ -30,6 +30,12 @@
   ]
   $purge_plugins = false
 
+  if versioncmp(pick($facts['jenkins_version'], '0'), '2.313') >= 0 {
+    $systemd_type = 'simple'
+  } else {
+    $systemd_type = 'forking'
+  }
```
ekohl, line 37: Wouldn't it be safer to assume `simple` and only if the fact is an older version to return `forking`?

During the initial run the fact won't be available (since the war is not present) so you'll always need 2 runs.

I would also be happy to skip automatic detection and require the admin to select `forking` if they're on an older version. Even the current LTS should default to simple and they don't support older versions. This saves a lot of computing power (starting up Java just to get a version number) that most people won't need.

ekohl (changes requested):
I left a bunch of comments in the fact but I'm not convinced we really need a fact. See the comment in `params.pp` for that reasoning.

lib/facter/jenkins.rb
fe80: Bad c/c I guess

lib/facter/jenkins.rb
fe80: Np :)

manifests/params.pp
fe80: Yes during the initial setup that not work for the last release.

I can change like that if you want:
`if versioncmp(pick($facts['jenkins_version'], '2.313'), '2.313') >= 0 {`

That just create a breaking change if the libdir is changed and the fact not resolve

fe80: I just wrote a dynamic solution but we can remove the fact if you want

```
time /bin/java -jar /usr/lib/jenkins/jenkins.war --version
2.319.1

real	0m0.371s
user	0m0.562s
sys	0m0.040s
```

fe80: We can also remove the `--deamon` and set simple options for all jenkins versions

ekohl: I was actually thinking about swapping the logic so it's:
```suggestion
  if versioncmp(pick($facts['jenkins_version'], '0'), '2.313') < 0 {
    $systemd_type = 'forking'
  } else {
    $systemd_type = 'simple'
  }
```

That way it defaults to `simple` if the fact is not found.

However, in general I'm happier with looking to the future and aligning on `simple`, removing `forking` altogether. With that I will note that I haven't looked at this area so I may be missing something.

fe80: I haven't really watched either... In fact I don't understand why the package doesn't create the unit file and why in Debian family they use a init script... And I don't know if it's work with Debian family

For imformation `--deamon` option is set in the `/etc/sysconfig/jenkins` but it's useless since `2.313`

I can try on a Ubuntu 20.04 this week or next week If you want

ekohl: It's actually an excellent question and I wondered the same thing. Perhaps we should actually file an upstream bug. It may just be an oversight.

fe80: https://github.com/jenkinsci/packaging/issues/216

manifests/params.pp
```diff
@@ -30,6 +30,12 @@
   ]
   $purge_plugins = false
 
+  if versioncmp(pick($facts['jenkins_version'], '0'), '2.313') < 0 {
```
ekohl, line 33: Doesn't this still default to forking when the fact is missing? `'0' < '2.313'`, right? Could you add a test for this?

ekohl (commented):
I hit this issue myself when I upgraded to 2.313 today. I can confirm that I needed to change it to `Type=simple` to avoid a restart loop.

manifests/params.pp
fe80: like this it's ok for you ?

spec/classes/jenkins_service_spec.rb
```diff
@@ -76,6 +77,27 @@
           end
         end
       end
+      if os_facts[:os]['release']['major'].to_i >= 7
```
ekohl, line 80: I'd prefer to rebase this on https://github.com/voxpupuli/puppet-jenkins/pull/1026. That removes the need for this conditional.

templates/jenkins-run.erb
```diff
@@ -55,7 +55,9 @@ JAVA_CMD="$JENKINS_JAVA_CMD $JENKINS_JAVA_OPTIONS -DJENKINS_HOME=$JENKINS_HOME -
 PARAMS=()
 PARAMS+=("--logfile=/var/log/jenkins/jenkins.log")
 PARAMS+=("--webroot=/var/cache/jenkins/war")
+<%- if scope['jenkins::systemd_type'] == 'simple' -%>
```
ekohl, line 58: Shouldn't this be `forking`?
```suggestion
<%- if scope['jenkins::systemd_type'] == 'forking' -%>
```

Perhaps you could include a test for it in the `with jenkins version x.y` tests too.

spec/classes/jenkins_service_spec.rb
ekohl: This now merged. Could you rebase?

fe80: I thinks it's ok like that ?

ekohl (approved):
Thanks for working with me on this.

fe80:
Thanks to you for your help and your tips :+1:

PR #1027, opened 2021-12-14
Remove Optional[] where the param has a default

ekohl:
This is found by the optional_default check.


bastelfreak (approved):

bastelfreak (approved):

ekohl:
Looks like EL7 now fails because it requires daemonize, which I suspect is in EPEL. Let's ignore acceptance tests for now.

PR #1028, opened 2022-02-09
Don't put unparseable JSON in output

jvdmr:
Unparseable JSON may still contain valid credentials or other sensitive data that should not be added to logs in plaintext.

Remove the variable that possibly contains sensitive data from the error message when parsing fails.

n/a

### DIFF
--- a/lib/puppet/provider/jenkins_credentials/cli.rb
+++ b/lib/puppet/provider/jenkins_credentials/cli.rb
@@ -108,7 +108,7 @@ Puppet::Type.type(:jenkins_credentials).provide(:cli, parent: Puppet::X::Jenkins
     begin
       JSON.parse(raw)
     rescue JSON::ParserError
-      raise Puppet::Error, "unable to parse as JSON: {REDACTED}"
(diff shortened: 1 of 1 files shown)


lib/puppet/provider/jenkins_credentials/cli.rb
```diff
@@ -108,7 +108,7 @@ def self.credentials_list_json(catalog = nil)
     begin
       JSON.parse(raw)
     rescue JSON::ParserError
-      raise Puppet::Error, "unable to parse as JSON: #{raw}"
+      raise Puppet::Error, "unable to parse as JSON: {REDACTED}"
```
ekohl, line 111: I wonder if we should instead show the error. Capture the exception message and show that. Otherwise rephrase it as:
```suggestion
      raise Puppet::Error, 'Unable to parse Jenkins credentials list as JSON'
```

ekohl (commented):
This makes sense to me.

ekohl (approved):

PR #1029, opened 2022-02-28
Fix/jenkins credentials

zipkid:
When updating credentials or using the puppet debug flag credentials leak to the puppet log.
This PR is an attempt to avoid this without requiring the user to apply the Sensitive data type to all parameters.

### DIFF
--- a/lib/puppet/type/jenkins_credentials.rb
+++ b/lib/puppet/type/jenkins_credentials.rb
@@ -51,26 +51,152 @@ Puppet::X::Jenkins::Type::Cli.newtype(:jenkins_credentials) do
 
   newproperty(:password) do
     desc 'password - UsernamePasswordCredentialsImpl, CertificateCredentialsImpl'
+    def is_to_s(_current_value)
+      '[old password]'
+    end
+
+    def should_to_s(_new_value)
+      '[new password]'
+    end
+
+    def change_to_s(_current_value, _new_value)
+      begin
+        if _current_value == :absent
+          return "Added '#{name}' as #{should_to_s(_new_value)}"
+        elsif _new_value == :absent or _new_value == [:absent]
+          return "Removed '#{name}' from #{is_to_s(_current_value)}"
+        else
+          return "Changed '#{name}' from #{is_to_s(_current_value)} to #{should_to_s(_new_value)}"
+        end
+      rescue Puppet::Error
+        raise
+      end
+    end
   end
 
   newproperty(:private_key) do
     desc 'ssh private key string - BasicSSHUserPrivateKey'
+    def is_to_s(_current_value)
+      '[old ssh private key string]'
+    end
+
+    def should_to_s(_new_value)
+      '[new ssh private key string]'
+    end
+
+    def change_to_s(_current_value, _new_value)
+      begin
+        if _current_value == :absent
+          return "Added '#{name}' as #{should_to_s(_new_value)}"
+        elsif _new_value == :absent or _new_value == [:absent]
+          return "Removed '#{name}' from #{is_to_s(_current_value)}"
+        else
+          return "Changed '#{name}' from #{is_to_s(_current_value)} to #{should_to_s(_new_value)}"
+        end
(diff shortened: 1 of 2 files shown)


lib/puppet/type/jenkins_credentials.rb
```diff
@@ -51,26 +51,152 @@
...
+      begin
+        if _current_value == :absent
+          return "Added '#{name}' as #{should_to_s(_new_value)}"
+        elsif _new_value == :absent or _new_value == [:absent]
+          return "Removed '#{name}' from #{is_to_s(_current_value)}"
+        else
+          return "Changed '#{name}' from #{is_to_s(_current_value)} to #{should_to_s(_new_value)}"
+        end
+      rescue Puppet::Error
+        raise
```
ekohl, line 72: What's the point of this?

Also, since Ruby 2.5 you can write:
```ruby
def my_method
  # ...
rescue Puppet::Error
  # ...
end
```

zipkid: This was in the code i copied. Clearly unneeded so i removed it.

lib/puppet/type/jenkins_credentials.rb
```diff
@@ -51,26 +51,140 @@
...
+    def is_to_s(_current_value)
+      '[old password]'
+    end
+
+    def should_to_s(_new_value)
+      '[new password]'
+    end
+
+    def change_to_s(_current_value, _new_value)
+      begin
```
ekohl, line 63: You can also drop the `begin` and `end` then.

lib/puppet/x/jenkins/provider/cli.rb
```diff
@@ -130,7 +130,9 @@ def self.cli(command, options = {}, cli_pre_cmd = [])
                          false
                        end
 
-    Puppet.debug("#{sname} stdin:\n#{input}")
+    # This Puppet.debug shows the JSON with changed credentials in plain text.
+    # ToDo: Remove this debug output OR mask the credentials
+    # Puppet.debug("#{sname} stdin:\n#{input}")
```
ekohl, line 135: If this contains credentials, you should note that it's written to a file just below and there's a `chmod 644` rescue block, which makes those credentials world readable.

It looks like that is designed for the case where the `jenkins` user is not yet available. I wonder if that workaround should be removed and instead hard fail.

It should also be noted that there's no way to ensure the file really is removed. If `execute_with_retry` raises an exception it's not removed. That means in some cases the credentials could remain world readable until `$TMPDIR` is cleaned.

zipkid: At least this would stay on the server and not be shipped off to something like The Foreman where many more can see the credentials.

root-expert:
Hello,

I'm not a fan of this solution, that places a lot of code in the provider to fix a issue that there's already a solution for it.
IMHO if the user wants to hide credentials/sensitive info from leaking to the logs, then they should use the proper solution, Sensitive data type, instead of relying to hacky workarounds.

If, for some reason, the module doesn't support Sensitive data types that's something that needs to be added instead of using workarounds.

zipkid:
We looked into solving this via the sensitive datatype.

Because of the multiple implementation types and the many different credentials we use in Jenkins this is very impractical.
For each implementation type we need a different hiera sensitive setting.
Because we pass all credentials in a hash that requires a setting for each var.

This 'solution' is works for everyone & always, also for people who forget about or don't know the sensitive data type.

I understand your objection and would agree if this was just 1 password.

Regards,

Stefan Goethals.

kenyon:
@zipkid you can use regexp in the `lookup_options`, would that work for converting to `Sensitive`? Might have to arrange your data structure differently, not sure how it looks. https://puppet.com/docs/puppet/7/hiera_merging.html#setting_lookup_options_to_refine_the_result_of_a_lookup-lookup-options-format

zipkid:
@kenyon We have tried, believe me. The problem is that sensitive does not work in hiera on members of a hash.

PR #1032, opened 2022-03-21
fix libdir for systemd integration

### LINKED ISSUE #1031: Does no longer work with jenkins 2.332.1 or 2.335 onwards
Since jenkins moved to systemd recently, there is a collision with the systemd implementation here.

fe80:
Hello,

This is a proposal, I probably need to fix the rspec before this fix was validate.

`/usr/lib/jenkins` doesn't exist anymore in the package, the war file is move on `/usr/share/java` (libdir variable). But this path is use for put some file require by Puppet `jenkins-cli.jar`, `jenkins-run`, `puppet_helper.groovy`. Maybe It necessary to move that.

* Change jenkins_version fact for multiple war location
* Change libdir with new value since jenkins use systemd (/usr/share/java)

Fixes #1031

### DIFF
--- a/lib/facter/jenkins.rb
+++ b/lib/facter/jenkins.rb
@@ -17,20 +17,26 @@ Facter.add(:jenkins_version) do
   confine kernel: 'Linux'
 
   setcode do
-    libdir = case Facter.value('os.family')
-             when 'Debian'
-               '/usr/share/jenkins'
-             when 'Archlinux'
-               '/usr/share/java/jenkins'
-             else
-               '/usr/lib/jenkins'
-             end
-    war = libdir + '/jenkins.war'
+    libdirs = [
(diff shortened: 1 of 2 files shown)


fe80:
Effectively @sbeaulie 

If the systemd overwrite is remove, we need to review some params implementation, that very not the same change and create a breaking change for (not very) old jenkins releases.

The systemd support is a new feature from jenkins packaging re-up with https://github.com/voxpupuli/puppet-jenkins/pull/1025#discussion_r763884917

ekohl:
I started with this in https://github.com/voxpupuli/puppet-jenkins/pull/1035, but it's not ready yet.

ekohl:
Made obsolete by #1035

PR #1033, opened 2022-03-23
support multiple arguments in JENKINS_ARGS separated by spaces

sbeaulie:
before this change setting mutiple arguments like --sessionTimeout=480 --sessionEviction=28800
would fail when starting jenkins with errors similar to `java.lang.NumberFormatException: For input string: "480 --sessionEviction=28800"`
this means we need to separate each arg and load them separately in the PARAMS bash array

N/A


ekohl (changes requested):
The whole jenkins-run is no longer needed and shipped by the latest jenkins version (which includes the systemd unit file as well). I think it'd be better to make sure that can be used instead.

ekohl:
Made obsolete by #1035

PR #1035, opened 2022-03-28
Rewrite to native systemd

ekohl:
This takes a stab at rewriting to use Jenkins' 2.332 native systemd file. While writing this I found out several other issues with the test suite. It still isn't passing, but this is submitted to share my progress.

Several changes can be merged already and can be split off.


sbeaulie:
Are you planning to support both versions <2.332 (with module's systemd) and >2.332 (with jenkins packaged systemd) or would we release this module only for the newer jenkins versions?

ekohl:
I think just release it for the newer Jenkins. Perhaps release it as a new major release after such a long time.

ekohl:
(last push was just a trivial rebase)

bastelfreak (approved):

ekohl:
Includes https://github.com/voxpupuli/puppet-jenkins/pull/1038. The test suite isn't green yet, mostly because of plugins. In particular, the latest credentials plugin requires a newer-than-LTS version which breaks a lot of tests.

ekohl:
Thinking more about this, I'm leaning to switching CI to non-LTS for now.

ekohl:
I added sufficient workarounds that I now expect the tests to pass. However, I didn't run them locally so no guarantees.

ekohl:
Ok, the non-LTS fix I tried doesn't work because the modulesync is too old. I'm going to merge this as is now and then do a fresh modulesync.

PR #1036, opened 2022-03-28
More puppet-strings conversion

ekohl:
This isn't complete, but it makes progress.

### DIFF
--- a/manifests/augeas.pp
+++ b/manifests/augeas.pp
@@ -1,40 +1,43 @@
-# == Definition jenkins::augeas
+# @summary Change config files using augeas
 #
-# === Parameters:
+# @param config_filename
+#   Filename of the configuration file to work on relative to the jenkins
+#   localstatedir.
 #
-# [*config_filename*]:  Filename of the configuration file to work on relative to the jenkins localstatedir.
+# @param changes
+#   String or array with augeas changes to perform.
 #
-# [*changes*]:          String or array with augeas changes to perform.
+# @param onlyif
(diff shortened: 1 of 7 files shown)


manifests/augeas.pp
```diff
@@ -1,40 +1,43 @@
-# == Definition jenkins::augeas
+# @summary Change config files using augeas
 #
-# === Parameters:
+# @param config_filename
+#   Filename of the configuration file to work on relative to the jenkins
+#   localstatedir.
```
bastelfreak, line 5: I prefer it without a linebreak within the description, but as long as puppet-strings renders it properly I'm happy.

manifests/credentials.pp
```diff
@@ -12,10 +12,7 @@
 # See the License for the specific language governing permissions and
```
bastelfreak, line 12: should we keep this text or maybe move it to `@summary`?

ekohl: I wasn't quite sure. Perhaps we can just drop it altogether since the whole module is already licensed.

manifests/augeas.pp
ekohl: I always wrap my parameter documentation to a reasonable line length. In this case it may be more readable if I don't, but if there are more words ...

bastelfreak (approved):
I think it's a good first step. I'm fine with merging it now if you don't have the time to improve it in the next days. rebasing such changes is always annoying.

PR #1038, opened 2022-05-03
Drop sysvinit support in jenkins::slave

ekohl:
For the OS families where a template was provided now default to using systemd and we don't test with sysvinit.

This drops the ability to smoothly transition from sysvinit to systemd, but that's not really important.  Very few users will switch their system from sysvinit to systemd and if they do, it's a bigger upgrade anyway.

It doesn't implement any cleanup since it's not expected to be present.

It also corrects thje SELinux type for jenkins-slave-run which was needed to get it to run on EL8 with a real VM.


ekohl:
The current test suite is broken, but this is a step towards fixing that.

ekohl:
At least the Ruby tests should be green. Rebasing on https://github.com/voxpupuli/puppet-jenkins/pull/1039.

PR #1041, opened 2022-05-05
Fix acceptance tests

ekohl:
This is incomplete, see the last commit message for details. Sharing here in case anyone has time to dig into it.


ekohl:
The error during setup is completely unexpected. I'm not sure where that came from.

ekohl:
So it looks like on Ubuntu there is an initial user account: an admin user is created with the password in `/var/lib/jenkins/secrets/initialAdminPassword`. Probably because on Debian a service is started automatically while on RPM it isn't. However, the entire test suite depends on anonymous admin access. I'm leaning to merging this now.

ekohl:
I did some testing on Ubuntu 20.04. We do pass the systemd override:
```ini
# /etc/systemd/system/jenkins.service.d/puppet-overrides.conf
[Service]
Environment="JAVA_OPTS=-Djava.awt.headless=true -Djenkins.install.runSetupWizard=false"
```

However, then:
```console
# systemctl status jenkins.service 
● jenkins.service - Jenkins Continuous Integration Server
     Loaded: loaded (/lib/systemd/system/jenkins.service; enabled; vendor preset: enabled)
    Drop-In: /etc/systemd/system/jenkins.service.d
             └─puppet-overrides.conf
     Active: active (running) since Mon 2022-05-16 11:35:10 UTC; 23min ago
   Main PID: 9526 (java)
      Tasks: 38 (limit: 1066)
     Memory: 284.3M
     CGroup: /system.slice/jenkins.service
             └─9526 /usr/bin/java -Djava.awt.headless=true -jar /usr/share/java/jenkins.war --webroot=/var/cache/jenkins/war --httpPort=8080
```

Note how it is not respected. I then modified it and did a daemon-reload and it was there. I then suspected the problem is with Debian starting the daemon by default on installation (`apt install jenkins`) and then Puppet comes in and configures things. However, after a fresh install I tried a simple `systemctl restart jenkins` and it was insufficient. I needed an explicit `systemctl daemon-reload`. Then a restart did pick it up. I think I've seen this before with systemd on Ubuntu 20.04: it didn't pick up that /etc/systemd/system/jenkins.service.d was created.

I did notice a refresh from the drop in to the service was missing. After adding that I see:
```console
# systemctl status jenkins.service
Warning: The unit file, source configuration file or drop-ins of jenkins.service changed on disk. Run 'systemctl daemon-reload' to reload units.
● jenkins.service - Jenkins Continuous Integration Server
     Loaded: loaded (/lib/systemd/system/jenkins.service; enabled; vendor preset: enabled)
     Active: active (running) since Mon 2022-05-16 12:22:19 UTC; 28s ago
   Main PID: 9541 (java)
      Tasks: 48 (limit: 1066)
     Memory: 230.9M
     CGroup: /system.slice/jenkins.service
             └─9541 /usr/bin/java -Djava.awt.headless=true -jar /usr/share/java/jenkins.war --webroot=/var/cache/jenkins/war --httpPort=8080
```

So now it is detecting it at least. Question is why Puppet isn't detecting this. It does run `/usr/bin/systemctl show --property=NeedDaemonReload -- jenkins` as can be seen from the debug logs:

```
  Notice: /Stage[main]/Jenkins::Config/Systemd::Dropin_file[puppet-overrides.conf]/File[/etc/systemd/system/jenkins.service.d]/ensure: created
  Debug: /Stage[main]/Jenkins::Config/Systemd::Dropin_file[puppet-overrides.conf]/File[/etc/systemd/system/jenkins.service.d]: The container Systemd::Dropin_file[puppet-overrides.conf] will propagate my refresh event
  Notice: /Stage[main]/Jenkins::Config/Systemd::Dropin_file[puppet-overrides.conf]/File[/etc/systemd/system/jenkins.service.d/puppet-overrides.conf]/ensure: defined content as '{md5}dbb7a53b475d1a75b951997319ba96cb'
  Info: /Stage[main]/Jenkins::Config/Systemd::Dropin_file[puppet-overrides.conf]/File[/etc/systemd/system/jenkins.service.d/puppet-overrides.conf]: Scheduling refresh of Service[jenkins]
  Debug: /Stage[main]/Jenkins::Config/Systemd::Dropin_file[puppet-overrides.conf]/File[/etc/systemd/system/jenkins.service.d/puppet-overrides.conf]: The container Systemd::Dropin_file[puppet-overrides.conf] will propagate my refresh event
  Debug: Systemd::Dropin_file[puppet-overrides.conf]: The container Class[Jenkins::Config] will propagate my refresh event
  Debug: Class[Jenkins::Config]: The container Stage[main] will propagate my refresh event
  Info: Class[Jenkins::Service]: Scheduling refresh of Service[jenkins]
  Debug: Executing: '/usr/bin/systemctl is-active -- jenkins'
  Debug: Executing: '/usr/bin/systemctl is-enabled -- jenkins'
  Debug: Executing: '/usr/bin/systemctl is-active -- jenkins'
  Debug: Executing: '/usr/bin/systemctl show --property=NeedDaemonReload -- jenkins'
  Debug: Executing: '/usr/bin/systemctl restart -- jenkins'
  Notice: /Service[jenkins]: Triggered 'refresh' from 2 events
```

Could this be a race condition? As I said: I'm pretty sure I've seen this before on Ubuntu 20.04 but I'm not sure where to look further.

Even when we do fix this I suspect we'll run in the problem that the daemon is started before the override is created. For that a possible solution might be [systemd.preset](https://www.freedesktop.org/software/systemd/man/systemd.preset.html) to disable auto start, allowing Puppet to configure things before it comes up.

tl;dr: I do think this is a regression I introduced and probably blocks releasing a new module version.

ekohl:
So I pushed the notify fix. This doesn't solve it, but it's a bugfix nonetheless.



> Could this be a race condition? As I said: I'm pretty sure I've seen this before on Ubuntu 20.04 but I'm not sure where to look further.

The test confirms that it's not an issue on Ubuntu 18.04 and 20.04 isn't in metadata.json as supported, so we can consider that issue out of scope for now. The acceptance tests do find the issue on 20.04 so it'll be found when it is added.

bastelfreak (approved):

bastelfreak (approved):

ekohl:
The CentOS 7 failure looks like an intermittent network error:
```
  Error: /Stage[main]/Main/Jenkins::Plugin[apache-httpcomponents-client-4-api]/Archive[apache-httpcomponents-client-4-api.hpi]/ensure: change from 'absent' to 'present' failed: Execution of '/opt/puppetlabs/puppet/bin/curl https://updates.jenkins.io/latest/apache-httpcomponents-client-4-api.hpi -o /tmp/apache-httpcomponents-client-4-api.hpi_20220516-12838-g3hwbh -fsSLg --max-redirs 5' returned 28: curl: (28) Failed to connect to get.jenkins.io port 443: Connection timed out
```

Why Ubuntu isn't idempotent in downgrades is a mystery now. I'm not sure where those plugins come from.

ekohl:
> Why Ubuntu isn't idempotent in downgrades is a mystery now. I'm not sure where those plugins come from.

I think I solved that by increasing the version number of the jquery3-api plugin.

ekohl:
It is :green_apple:, quick merge it before it breaks again ;)

bastelfreak (approved):

PR #1044, opened 2022-05-18
Drop-in file systemd configuration

### LINKED ISSUE #1009: Add a way to set the open files limit in systemd service
Hello,

The sysconfig file contains a variable (MAXOPENFILES) that is used to set the limit for the opened files.

The problem is that under RHEL7, the hard limit set by systemd is 4096, that means that nothing can change that value to an higher value.

The .service file should be modified to allow setting this to an higher value (with probably the same default as the one in sysconfig, 8192)

jan-win1993:

Currently the Systemd drop-in file can only override environment variables.
I would like to expand this functionality to make it possible to override every configuration in the **[Service]** section of the drop-in file.

Fixes #1009

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -64,6 +64,9 @@
 #   * ``running`` (default)
 #   * ``stopped``
 #
+# @param service_override
+#   Override the jenkins service configuration
+#
 # @param service_provider
 #   Override ``Service[jenkins]`` resource provider
 #
@@ -281,6 +284,7 @@ class jenkins (
   Boolean $manage_service                         = true,
   Boolean $service_enable                         = true,
   Enum['running', 'stopped'] $service_ensure      = 'running',
+  Hash[String[1], String] $service_override       = {},
   Optional[String] $service_provider              = undef,
   Hash $config_hash                               = {},
   Hash $plugin_hash                               = {},
--- a/templates/jenkins-override.epp
+++ b/templates/jenkins-override.epp
@@ -1,5 +1,10 @@
-<%| Hash[String, Struct[{value => Any}]] $environment |-%>
+<%- | Hash[String, Struct[{value => Any}]] $environment,
+      Hash[String[1], String] $dropin_config = {}
+| -%>
 [Service]
 <% $environment.each |$key, $entry| { -%>
 Environment="<%= $key %>=<%= $entry['value'] %>"
+<% } -%>
+<% $dropin_config.each |$key, $value| { -%>
+<%= $key %>=<%= $value %>
 <% } -%>
--- a/manifests/user_setup.pp
+++ b/manifests/user_setup.pp
@@ -62,7 +62,7 @@ class jenkins::user_setup {
 
   systemd::dropin_file { 'puppet-overrides.conf':
     unit           => 'jenkins.service',
-    content        => epp("${module_name}/jenkins-override.epp", { 'environment' => $config_hash }),
+    content        => epp("${module_name}/jenkins-override.epp", { 'environment' => $config_hash, 'dropin_config' => $jenkins::service_override }),
     notify_service => true,
   }
 }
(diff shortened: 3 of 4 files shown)


manifests/init.pp
```diff
@@ -281,6 +284,7 @@
   Boolean $manage_service                         = true,
   Boolean $service_enable                         = true,
   Enum['running', 'stopped'] $service_ensure      = 'running',
+  Optional[Hash] $service_override                = undef,
```
ekohl, line 287: I'd simplify this to:
```suggestion
  Hash $service_override                = {},
```

You'll need to modify all other instances as well, but this simplifies things. Also, the type in the template itself is stricter than what you have here. I'd make this the same so puppet-strings can properly extract it.

templates/jenkins-override.epp
```diff
@@ -1,5 +1,12 @@
-<%| Hash[String, Struct[{value => Any}]] $environment |-%>
+<%- | Hash[String, Struct[{value => Any}]] $environment,
+      Optional[Hash[String, Struct[{value => Any}]]] $dropin_config = undef
```
ekohl, line 2: Do you expect to place entries here? Or why do you have the struct as a value with just a single entry (`value`)?

I think for the env vars I kept it to the complex type (with `value`) out of backwards compatibility for the parameter but for a new parameter I'd consider making it simple.

ekohl (commented):
It looks like all tests fail. I'm not sure if that's related.

ekohl:
Please rebase. I just merged https://github.com/voxpupuli/puppet-jenkins/pull/1041 which means the tests should be green.

templates/jenkins-override.epp
```diff
@@ -1,5 +1,12 @@
-<%| Hash[String, Struct[{value => Any}]] $environment |-%>
+<%- | Hash[String, Struct[{value => Any}]] $environment,
+      Optional[Hash[String, String]] $dropin_config = undef
```
ekohl, line 2: I'd also make this a Hash
```suggestion
      Hash[String, String] $dropin_config = {},
```

templates/jenkins-override.epp
```diff
@@ -1,5 +1,12 @@
...
+| -%>
 [Service]
 <% $environment.each |$key, $entry| { -%>
 Environment="<%= $key %>=<%= $entry['value'] %>"
 <% } -%>
+<% unless $dropin_config =~ Undef { -%>
+<% $dropin_config.each |$key, $value| { -%>
+<%= $key %>=<%= $value %>
+<% } -%>
+<% } -%>
```
ekohl, line 12: Then the `Undef` line becomes redundant
```suggestion
<% $dropin_config.each |$key, $value| { -%>
<%= $key %>=<%= $value %>
<% } -%>
```

manifests/init.pp
```diff
@@ -281,6 +284,7 @@
   Boolean $manage_service                         = true,
   Boolean $service_enable                         = true,
   Enum['running', 'stopped'] $service_ensure      = 'running',
+  Hash $service_override                          = {},
```
ekohl, line 287: If the EPP is `Hash[String, String]`, I'd make it the same here. I'd also consider `String[1]` to only allow non-empty strings, at least for the keys.

ekohl (changes requested):
It would be nice to have a test for this. Either a unit test or an acceptance test (note that's an `or`, not an `xor`).

templates/jenkins-override.epp
jan-win1993: Sorry, I am kind of a Puppet newbie. Thanks for the recommendations.
I will take this approach.

ekohl: No worries, we all had to start somewhere :)

ekohl (approved):
The code looks good to me. I just wonder if the test failures are related or my fixes were insufficient. I've opened https://github.com/voxpupuli/puppet-jenkins/pull/1045 which should give some answers.

PR #1046, opened 2022-06-23
fix: typo

Coukaratcha:


ekohl:
This is not a typo, but it's Ruby syntax that was introduced in Ruby 2.3: https://www.rubyinrails.com/2017/11/17/safe-navigation-operator-ampersand-dot-in-ruby/. It suggests you're on a very old Ruby and like an unsupported Puppet version.

Coukaratcha:
You're right. Sorry for this wrong PR. Thank you for the explanation.

ekohl:
No worries :)

PR #1047, opened 2022-07-28
add download option to jenkins module

CyberLine:
While on Debian10 plugin installation very often fails with:

`Error: /Stage[main]/Jenkins::Plugins/Jenkins::Plugin[conditional-buildstep]/Archive[conditional-buildstep.hpi]/ensure: change from 'absent' to 'present' failed: Execution of '/usr/bin/curl https://updates.jenkins.io/latest/conditional-buildstep.hpi -o /tmp/conditional-buildstep.hpi_20220728-1598563-1n3z4vb -fsSLg --max-redirs 5' returned 16: curl: (16) Error in the HTTP2 framing layer`

i added http1.1 hard as download option to prevent those errors.


manifests/plugin.pp
```diff
@@ -56,6 +59,7 @@
   Boolean $enabled                  = true,
   String $digest_type               = 'sha1',
   Boolean $pin                      = false,
+  Array $download_options           = ['--http1.1'],
```
ekohl, line 62: Should this really be the default?

CyberLine: Actually the error is depending on which server behind update.jenins.io is serving the request. I think there is no downside for now to always use http1.1 instead of 2.0?

ekohl: HTTP 2.0 can be more efficient. Have you at least filed an issue with Jenkins to track it? IMHO such a workaround always needs an issue and link it in a code comment.

CyberLine: No, there are a few servers behind and one of them is randomly having issues. i still haven't found out which server.
For me it will also be fine to remove the default but make it injectable via parameter so i can override it.

CyberLine: Ok. Got one...

`/usr/bin/curl --head https://updates.jenkins.io/latest/ansicolor.hpi -fsSLg --max-redirs 5; echo $?`

```
HTTP/2 302
date: Thu, 28 Jul 2022 12:26:52 GMT
server: Apache
location: https://updates.jenkins.io/download/plugins/ansicolor/1.0.2/ansicolor.hpi
content-type: text/html; charset=iso-8859-1

curl: (16) Error in the HTTP2 framing layer
16
```

CyberLine: https://github.com/jenkins-infra/helpdesk/issues/3078

CyberLine: Well ... https://github.com/jenkins-infra/helpdesk/issues/3078#issuecomment-1198084336

ekohl: Ok, that certainly helps in giving me assurances that this is the right thing to do.

ekohl: > And why the heck are the tests failing since ages?

It has been very hard to keep them green, mostly because we don't have proper dependency resolution. Ideally we'd have something that utilizes https://github.com/jenkinsci/plugin-installation-manager-tool/ so you don't need to do that yourself.

CyberLine: As i can see in the documentation, the tool also won't resolve dependencies for you :(

manifests/job.pp
```diff
@@ -13,7 +13,7 @@
 define jenkins::job (
   String $config,
   Optional[String] $source                  = undef,
-  Optional[Stdlib::Absolutepath] $template  = undef,
+  Optional[String] $template                = undef,
```
ekohl, line 16: Could you submit this as a separate PR?

ekohl (changes requested):
It took me a while, but I've finally found time to fix the tests. Please rebase.

manifests/job.pp
ekohl: Fixed in 697193012024c6cf8686dc0b29e018a57ab7a48a

ekohl:
I'm sorry I missed the rebase, and now there's another conflict. I'm also running into this, so I opened https://github.com/voxpupuli/puppet-jenkins/pull/1073 as a clean rebase with a better commit message.

evgeni:
#1073 was merged

PR #1064, opened 2023-01-31
el repo enabled toggle added

wimkorevaar:
Added a toggle to enable the repo in EL

### DIFF
--- a/manifests/repo.pp
+++ b/manifests/repo.pp
@@ -3,6 +3,7 @@
 class jenkins::repo (
   Stdlib::Httpurl $base_url = 'https://pkg.jenkins.io',
   String $gpg_key_filename = 'jenkins.io.key',
+  Integer $enabled = 1,
 ) {
   assert_private()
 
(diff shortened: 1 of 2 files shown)


manifests/repo.pp
```diff
@@ -3,6 +3,7 @@
 class jenkins::repo (
   Stdlib::Httpurl $base_url = 'https://pkg.jenkins.io',
   String $gpg_key_filename = 'jenkins.io.key',
+  Integer $enabled = 1,
```
ekohl, line 6: Pretty sure `yumrepo` also accepts a `Boolean`, which I think is cleaner.

ekohl (commented):
Can this parameter also apply to Debian and Suse?

manifests/repo.pp
wimkorevaar: Of course, just a brainfog moment.

wimkorevaar: I'm not that familiar with suse/debian, but I checked the .pp and currently no enable variable was available, but if it is possible to add the toggle to zypprepo or apt::source I won't be against it.

wimkorevaar:
The errors that are generated don't seem to be related to my PR in anyway.

ekohl (approved):
We can ignore the errors. Thanks!

PR #1069, opened 2023-04-12
Update Debian repository key for 2023

flichtenheld:
Update Debian repository key

See https://www.jenkins.io/doc/upgrade-guide/2.387/#upgrading-to-jenkins-lts-2-387-2

Note that this is really not related to 2.387.2 specifically since the whole stable repo is signed by the new key.


ekohl (approved):
Thanks! I ran into this for RPM, but the blog says Debian has the same thing.

PR #1076, opened 2023-07-01
Allow puppetlabs/stdlib 9.x, puppetlabs/java 10.x, puppet/archive 7.x, puppet/zypprepo 5.x, puppet/systemd 6.x

bastelfreak:

PR #1078, opened 2023-09-04
Allow puppet/systemd 5.x and 6.x

evgeni:
5.x dropped Puppet 6 support, which we don't support anyway
6.x dropped Ubuntu 18.04 support, which we should drop soon


bastelfreak (approved):

evgeni:
#1076 now contains it, as otherwise shit just explodes.

PR #1079, opened 2023-09-27
update plugins needed for integration tests

evgeni:


bastelfreak (approved):

evgeni:
You're quick with approving, but it's not working :/

PR #1084, opened 2023-11-23
Remove legacy top-scope syntax

smortex:


evgeni:
I don't think the changelog and history should be edited.

smortex:
> I don't think the changelog and history should be edited.

I adjusted the issues/PR titles and regenerated these files when I created this PR, so these changes will not be lost on the next release.  Is it what you are worried about, or do you prefer we revert the titles to re-add the legacy syntax and keep it as it is now?

evgeni (approved):

PR #1086, opened 2024-01-10
Cloudbees folder patch 1

mthirdy:
Compatibility with plugin cloudbees-folder'

This PR provides the ability to manage elements of type/plugin 'cloudbees-folder'. In this case the name of the job could include slashes for subfolders. So the slashes in the job name have to be replaced by underscore for the temporary created file used for the jenkins api. Also the slashes have to replaced by '/jobs/' for the config.xml of a job while deleting the job via the jenkins api.


mthirdy:
@evgeni what did you do, so the checks have passed?

evgeni:
Fixed up the tests in master and then restarted your PR against that

PR #1093, opened 2024-03-27
Fix acceptance test Java version

zilchms:
includes: #1092


bastelfreak (approved):

zilchms:
Ahhhh, now I understand. The acceptance tests have never worked in the first place. Well time to fix that then (or write new tests)

Valantin:
If is usefull I use this to install java 21 on my test jenkis setup on a ubuntu 22.04

```
java::package: openjdk-21-jdk
java::alternative: java-1.21.0-openjdk-amd64
java::alternative_path: /usr/lib/jvm/java-1.21.0-openjdk-amd64/bin/java
java::java_home: /usr/lib/jvm/java-1.21.0-openjdk-amd64/
```

I haven't checked if everything is necessary

spec/acceptance/hieradata/common.yaml
```diff
@@ -1,3 +1,3 @@
 ---
```
evgeni, line 1: the rename seems wrong, the package name is not common but specific to the RedHat family

spec/acceptance/hieradata/os/Debian.yaml
```diff
@@ -0,0 +1,3 @@
+---
+# Jenkins 2.361.1 requires Java 11 or newer
+java::package: openjdk-17-jdk
```
evgeni, line 3: you say 11, but use 17? :)

spec/acceptance/hieradata/common.yaml
ekohl: I was going to say that https://github.com/voxpupuli/voxpupuli-acceptance/?tab=readme-ov-file#hiera shows the default hierachy, but this module still uses the old voxpupuli-acceptance 2.x where `os` really is `family`. See https://github.com/voxpupuli/voxpupuli-acceptance/commit/650b6428205de4a01d6de0fd5e71881b7cfec7a0

evgeni: it does not since yesterday :)

evgeni:
#1115 did this and is merged

PR #1098, opened 2024-04-23
Credential plugins CI fixes

evgeni:


ekohl (approved):

evgeni:
I've split out the AWS and SSH part into https://github.com/voxpupuli/puppet-jenkins/pull/1099, as that works…

PR #1100, opened 2024-04-25
add test for proxy config

evgeni:


ekohl:
Shouldn't this fail on https://github.com/voxpupuli/puppet-jenkins/pull/1083?

evgeni:
That was my expectation, but after reading Jenkins source (ugh), I think that only happens when you save once in the UI, which I didn't test yet

PR #1103, opened 2024-04-29
Don't notify service when not managing it

yuri-moens:
The `jenkins::service` class is not included when `$jenkins::manage_service` is false. We should not try to restart it in the `proxy` class if we're not managing it.

### DIFF
--- a/manifests/proxy.pp
+++ b/manifests/proxy.pp
@@ -20,9 +20,14 @@ class jenkins::proxy {
       mode    => '0644',
     }
 
-    Package['jenkins']
-    -> File[$proxy_xml]
-    ~> Class['jenkins::service']
+    if $jenkins::manage_service {
+      Package['jenkins']
+      -> File[$proxy_xml]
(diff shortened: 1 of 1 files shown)


manifests/proxy.pp
```diff
@@ -20,9 +20,14 @@
...
-    -> File[$proxy_xml]
-    ~> Class['jenkins::service']
+    if $jenkins::manage_service {
+      Package['jenkins']
+      -> File[$proxy_xml]
+      ~> Class['jenkins::service']
+    } else {
+      Package['jenkins']
+      -> File[$proxy_xml]
+    }
```
smortex, line 30: What about splitting?

```suggestion
    Package['jenkins'] -> File[$proxy_xml]

    if $jenkins::manage_service {
      File[$proxy_xml] ~> Class['jenkins::service']
    }
```

Or maybe moving the `if $jenkins::manage_service` into the `jenkins::service` class and always notifying the class that may contain 0 resources? :shrug:

PR #1106, opened 2024-05-07
CentOS: Drop EoL 7/8 support

bastelfreak:
Both are soonish EoL so we should drop them.


evgeni:
Is OL7 als EOL then?

evgeni:
https://endoflife.date/oracle-linux says 31 Dec 2024

Well then

evgeni (approved):

PR #1112, opened 2024-09-20
Drop EoL EL7 support

bastelfreak:

### DIFF
--- a/manifests/plugin.pp
+++ b/manifests/plugin.pp
@@ -59,7 +59,7 @@ define jenkins::plugin (
   Boolean $enabled                   = true,
   String $digest_type                = 'sha1',
   Boolean $pin                       = false,
-  Array[String[1]] $download_options = ($facts['os']['family'] == 'RedHat' and $facts['os']['release']['major'] == '7') ? { true => [], default => ['--http1.1'] },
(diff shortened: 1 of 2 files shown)


rwaffen (approved):

REFERENCE.md
evgeni, line 1: why is this file here? shouldn't it be just generated on module release and "gut is"?

bastelfreak: during the release process, we've logic in place to verify that the reference.md is up to date, but it needs to exist.

evgeni: :thinking: ok

but still, this is "Drop EL7", it doesn't touch anything release-worthy.

Should that be part of https://github.com/voxpupuli/puppet-jenkins/pull/1111?

(yes, nitpicking, sorry)

ekohl: In Foreman we generate it as part of the release because IMHO we shouldn't include generated files in git. In Vox people did want that and I didn't have the energy to push back on it though I still hate it

ekohl: Can you at least make it a separate commit?

evgeni: I moved that comit over to #1111 (people really should not give me push permissions everywhere :stuck_out_tongue_closed_eyes: )

PR #1114, opened 2024-12-18
refactor relates to apt 10x release

flepoutre:
Hello,
Please find a minor update relates to the new apt release module.

I would also like to take this opportunity to remind you of this issue, the answer to which I did not really understand : https://github.com/voxpupuli/puppet-jenkins/issues/1110

Many thanks for your help.

### DIFF
--- a/manifests/repo/debian.pp
+++ b/manifests/repo/debian.pp
@@ -16,7 +16,7 @@ class jenkins::repo::debian (
   apt::source { 'jenkins':
     location => $location,
     release  => 'binary/',
-    repos    => '',
+    repos    => 'main',
     include  => {
       'src' => false,
     },
--- a/metadata.json
+++ b/metadata.json
@@ -79,7 +79,7 @@
     },
     {
       "name": "puppetlabs/apt",
-      "version_requirement": ">= 4.4.0 < 10.0.0"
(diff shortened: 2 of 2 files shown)


metadata.json
```diff
@@ -79,7 +79,7 @@
     },
     {
       "name": "puppetlabs/apt",
-      "version_requirement": ">= 4.4.0 < 10.0.0"
+      "version_requirement": ">= 4.4.0 <= 10.0.0"
```
kenyon, line 82:
```suggestion
      "version_requirement": ">= 4.4.0 < 11.0.0"
```

kenyon (changes requested):

manifests/repo/debian.pp
```diff
@@ -16,7 +16,7 @@
   apt::source { 'jenkins':
     location => $location,
     release  => 'binary/',
-    repos    => '',
+    repos    => 'main',
```
bastelfreak, line 19: Is this now still compatible with older puppetlabs/apt releases, or should we raise the minimal version in metadata.json to 10x?

flepoutre: I don't see why it wouldn't be compatible.

evgeni:
#1116 was merged instead

PR #1116, opened 2025-01-06
Support puppetlabs/apt 10.x; require >= 9.2

evgeni:

### DIFF
--- a/manifests/repo/debian.pp
+++ b/manifests/repo/debian.pp
@@ -16,7 +16,6 @@ class jenkins::repo::debian (
   apt::source { 'jenkins':
     location => $location,
     release  => 'binary/',
-    repos    => '',
     include  => {
       'src' => false,
     },
--- a/metadata.json
+++ b/metadata.json
@@ -79,7 +79,7 @@
     },
     {
       "name": "puppetlabs/apt",
-      "version_requirement": ">= 4.4.0 < 10.0.0"
+      "version_requirement": ">= 9.2.0 < 11.0.0"
     },
     {
       "name": "puppetlabs/java",


manifests/repo/debian.pp
```diff
@@ -16,7 +16,7 @@
   apt::source { 'jenkins':
     location => $location,
     release  => 'binary/',
-    repos    => '',
+    repos    => undef,
```
ekohl, line 19: If this is undef, can't we simply leave it out?
```suggestion
```

alexjfisher, line 19: I don't think `undef` is allowed.
https://github.com/puppetlabs/puppetlabs-apt/blob/966d39cfe2cd85a5b481e67e8da51e2024b336d2/manifests/source.pp#L111

Maybe an empty array instead??

alexjfisher: In either case, that would result in `main` being used?? (The module default).  I'm guessing this isn't the intention.

evgeni: No, the intention is to get an empty string in the generated file. Won't an explicit undef do that?

ekohl: No, `undef` means "use whatever is defined" elsewhere I think. So it ends up being `main`.

ekohl: AFAIK `undef` is always allowed. It just means it ends up using the default.

ekohl: Also, I wonder if an array works: https://github.com/puppetlabs/puppetlabs-apt/pull/1167/files#r1904058136

evgeni: It does *not* use the default, no.

```
# cat lol.pp 
$location = 'https://pkg.jenkins.io/debian-stable'
$gpg_key_id = '63667EE74BBA1F0A08A698725BA31D57EF5975CA'

    apt::source { 'jenkins':
      location => $location,
      release  => 'binary/',
      repos    => undef,
      include  => {
        'src' => false,
      },
      key      => {
        'id'     => $gpg_key_id,
        'source' => "${location}/jenkins.io-2023.key",
      },
    }

# puppet apply lol.pp 
Notice: Compiled catalog for eb3b185efcfa.redhat.com in environment production in 0.10 seconds
Notice: /Stage[main]/Apt/File[sources.list]/ensure: created
Notice: /Stage[main]/Apt/File[preferences]/ensure: created
Notice: /Stage[main]/Apt/Package[gnupg]/ensure: created
Notice: /Stage[main]/Apt/Apt::Setting[conf-update-stamp]/File[/etc/apt/apt.conf.d/15update-stamp]/ensure: defined content as '{sha256}2e6eb1f5f20262bfc6b7dfb26a302f00b4ab5fee803abd9e07ad8378cce067d5'
Notice: /Stage[main]/Main/Apt::Source[jenkins]/Apt::Key[Add key: 63667EE74BBA1F0A08A698725BA31D57EF5975CA from Apt::Source jenkins]/Apt_key[Add key: 63667EE74BBA1F0A08A698725BA31D57EF5975CA from Apt::Source jenkins]/ensure: created
Notice: /Stage[main]/Main/Apt::Source[jenkins]/Apt::Setting[list-jenkins]/File[/etc/apt/sources.list.d/jenkins.list]/ensure: defined content as '{sha256}f853a9845679b71b1e95e04d20508cabc490f9ba393274037bda37f87429408a'
Notice: /Stage[main]/Apt::Update/Exec[apt_update]: Triggered 'refresh' from 1 event
Notice: Applied catalog in 4.54 seconds

# cat /etc/apt/sources.list.d/jenkins.list
# This file is managed by Puppet. DO NOT EDIT.
# jenkins
deb https://pkg.jenkins.io/debian-stable binary/
```

evgeni:
As shown in https://github.com/voxpupuli/puppet-jenkins/pull/1115, this works "as intended" (but yes, I don't like that api change in puppetlabs/apt either).

manifests/repo/debian.pp
ekohl: It might, but you're probably hitting this code: https://github.com/puppetlabs/puppetlabs-apt/blob/966d39cfe2cd85a5b481e67e8da51e2024b336d2/manifests/source.pp#L140-L144

Note `release` is `binary/` so `$_components` is just `$_release`, not `${_release} ${repos}`. I expect that if you change `release` to `binary` you will see it as `deb https://pkg.jenkins.io/debian-stable binary main`.

If not, it would violate the type `Variant[String[1], Array[String[1]]]` which doesn't allow `undef`.

evgeni: Ah. Yes indeed.

WHY IS THERE SO MUCH MAGIC?!

ekohl: I don't know, but I still think we should leave it out
```suggestion
```

evgeni: Done. Requires https://github.com/puppetlabs/puppetlabs-apt/commit/ce8e484230aad11339495613d7e46e7e4ecf00e9 tho.

ekohl (approved):

PR #1130, opened 2025-07-22
use apt::keyring to install the Jenkins APT repo key in Debian

### LINKED ISSUE #1129: apt::key doesn't work in Debian trixie → switch to apt::keyring instead
The `apt-key` tool is not present in Debian trixie and installing Jenkins from the upstream repository fails with:

```
Warning: /Stage[main]/Jenkins::Repo::Debian/Apt::Source[jenkins]/Apt::Key[Add key: 63667EE74BBA1F0A08A698725BA31D57EF5975CA from Apt::Source jenkins]/Anchor[apt_key 63667EE74BBA1F0A08A698725BA31D57EF5975CA present]: Skipping because of failed dependencies
```

i think this module needs to switch to `apt::keyring` instead.

zen-fu:

apt-key is not present in Debian Trixie, and because of that the `apt::key` resource doesn't work in that release.  Upstream [puppetlabs-apt](https://github.com/puppetlabs/puppetlabs-apt) advises:

```
Warning apt::key is deprecated in the latest Debian and
Ubuntu releases. Please use apt::keyring instead.
```

We also implement a check for the expected fingerprint, as that's not done by the upstream code (even though it probably should support that?). This way, we will at least get an error in Puppet if the fingerprint doesn't match.

Fixes #1129

### DIFF
--- a/manifests/repo/debian.pp
+++ b/manifests/repo/debian.pp
@@ -1,8 +1,6 @@
 # @summary Set up the apt repo on Debian-based distros
 # @api private
-class jenkins::repo::debian (
-  String $gpg_key_id = '63667EE74BBA1F0A08A698725BA31D57EF5975CA',
-) {
+class jenkins::repo::debian {
   assert_private()
 
   include apt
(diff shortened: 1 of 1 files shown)


kenyon:
My concern is that apt key verification functionality doesn't really belong in this module, it belongs in puppetlabs-apt. Unfortunately it seems like Perforce has given up on maintaining their modules.

smortex:
The point of @kenyon is right, this verification belongs to the apt module so that all consumers can benefit from it.  If Perforce fail at maintaining these modules, I guess that at the point we are currently, we can just fork them and start to ship puppet-apt.  Maybe something to discuss at the next vox meetup?

sboyd-m:
It's definitely worth looking at this again, as Jenkins has updated their keys. https://www.jenkins.io/blog/2025/12/23/repository-signing-keys-changing/

Frustratingly they only list the new key and don't seem to mention the ids of the old keys, but here are some:
`62A9756BFD780C377CF24BA8FCEF32E745F2C3D5` expired 2023
`63667EE74BBA1F0A08A698725BA31D57EF5975CA` will expire 26 Mar 2026 (deprecated)
`5E386EADB55F01504CAE8BCF7198F4B714ABFC68` will expire 21 Dec 2028 (current)

PR #1140, opened 2025-12-29
modulesync 10.5.0

pccibot:
modulesync 10.5.0

PR #1141, opened 2026-01-28
Add new GPG Keys for 2026 Release

C24-AK:
This Pull Request is adding the new Jenkins Key Ring which is stated here: https://pkg.jenkins.io/debian-stable/


kenyon:
We need #1130 and #1140.

PR #1143, opened 2026-03-16
Use `apt::keyring` and update GPG key, with modulesync changes

### LINKED ISSUE #1129: apt::key doesn't work in Debian trixie → switch to apt::keyring instead

griggi-ws:
This implements #1130, #1140, and #1141. 
I've regenerated the rubocop todo alongside the modulesync changes, as prior to them rubocop passes and the updates result in lots of new warnings (and the lack of the GPG update renders this module useless as-is).

There's additionally a new stub to fix the cli_spec test, as it currently passes or fails depending on whether the test environment has a `java` binary in PATH. This will fix the CI run failure on #1142. 

Fixes #1129


grant-veepshosting:
+1 on this approach. We independently hit the same problem on a ~250-node fleet (Jenkins apt key rotation breaking `apt::key`-based installs) and had been carrying a local patch switching the repo key to the downloaded-keyring form. While preparing to upstream it we found this PR — it covers everything our patch did and more (including the modulesync updates), so we've dropped ours in favour of this. Routing through `apt::keyring` rather than the deprecated `apt::key` also matches what puppetlabs-apt itself steers toward, and is the only path that works on Debian trixie (#1129).

Happy to test this branch on our fleet if that helps it land.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1151, opened 2026-07-07
Allow a custom CLI URL

### LINKED ISSUE #1144: set_slaveagent_port fails
Since upgrading to Jenkins v2.541.3, we are getting the following error

`
/Stage[main]/Jenkins/Jenkins::Cli::Exec[set_slaveagent_port]/Exec[set_slaveagent_port]/returns

change from 'notrun' to ['0'] failed: '/bin/cat /usr/share/java/puppet_helper.groovy | /usr/bin/java -jar /usr/share/java/jenkins-cli.jar -s http://127.0.0.1:8080 -auth 'jenkins:ohghiequi3Et' groovy = set_slaveagent_port 50000' returned 255 instead of one of [0] (corrective)
`

This appears to be as a result of the following issue / patch:

https://github.com/jenkinsci/jenkins/issues/26507

https://www.jenkins.io/security/advisory/2026-03-18/#SECURITY-3674

The puppet module uses 127.0.0.1 to set_slaveagent_port.
 `
     let(:helper_cmd) { '/bin/cat /usr/share/java/puppet_helper.groovy | /usr/bin/java -jar /usr/share/java/jenkins-cli.jar -s http://127.0.0.1:8080 groovy =' }
`
https://github.com/voxpupuli/puppet-jenkins/blob/master/spec/defines/jenkins_cli_exec_spec.rb#L11C136-L11C145

### LINKED ISSUE #1150: Management of resources via WebSocket CLI fails with Unexpected Request Origin after Jenkins 2.555 security update
Jenkins 2.555, LTS 2.541.3 uses the configured Jenkins URL (from Manage Jenkins » System) rather than HTTP request headers to compute the expected origin for comparison, and refuses CLI WebSocket connections if they don’t match or the Jenkins URL is not configured. [https://www.jenkins.io/security/advisory/2026-03-18/#SECURITY-3674](url)

The puppet-jenkins module uses a http://localhost:<port> for all of it's connections to do configuration, which will not match the URL that is configured in Jenkins that people access it on through a browser.  This results in an unexpected request origin error:

`/usr/bin/java -jar /usr/share/java/jenkins-cli.jar -s http://localhost:8080 -auth 'someuser:somepassword' somegroovy`

`CLI handshake failed with status code 403
Content-Length: 0
Content-Security-Policy-Report-Only: base-uri 'none'; default-src 'self'; form-action 'self'; frame-ancestors 'self'; img-src 'self' data:; script-src 'report-sample' 'self' usage.jenkins.io; style-src 'report-sample' 'self' 'unsafe-inline'; report-to content-security-policy; report-uri <some custom jenkins URL>/content-security-policy-reporting-endpoint/3JIqRlMEtZdZFaRKtfhTiGDXkPj_ZS7v3JPpZ7cTa0w=:enBqZW5raW5z::L2NsaS93cw==
Date: Mon, 06 Jul 2026 18:06:04 GMT
Reporting-Endpoints: content-security-policy: <some custom jenkins url>/content-security-policy-reporting-endpoint/3JIqRlMEtZdZFaRKtfhTiGDXkPj_ZS7v3JPpZ7cTa0w=:enBqZW5raW5z::L2NsaS93cw==
Server: Jetty(12.1.8)
Vary: Accept-Encoding
X-CLI-Error: Unexpected request origin (check your reverse proxy settings)
X-Content-Type-Options: nosniff
`

If the system config in Jenkins under Jenkins Location > Jenkins URL is changed to "http://localhost:8080/", the cli commands works without error.

pwillworth:
Updates the cli and cli_helper command construction to use a new jenkins class parameter for the base url connection rather than being hardcoded to the local loopback interface.  This allows users of the module to customize this to match what they have configured in their system Jenkins Location, which is needed in newer version of Jenkins which check for unexpected request origin.
To support this, also hooked up the existing "port" parameter to be looked at by the jenkins_port function so that can be customized also.

Fixes #1150 
Fixes #1144

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -199,7 +199,7 @@
 # @param urlbase
 #  Jenkins base url including protocol (http/https) and hostname
 #
-#  Note that this value is use for CLI communication and does not configure your Jenkins system location
+#  Note that this value is used for CLI communication and does not configure your Jenkins system location.
 #  It should however match the base of your Jenkins Location in system config to prevent CORS
 #  unexpected Origin errors.
 #
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -196,6 +196,13 @@
 # @param cli_try_sleep
 #   Seconds between tries to contact jenkins API
 #
+# @param urlbase
+#  Jenkins base url including protocol (http/https) and hostname
+#
+#  Note that this value is use for CLI communication and does not configure your Jenkins system location
+#  It should however match the base of your Jenkins Location in system config to prevent CORS
+#  unexpected Origin errors.
+#
 # @param port
 #   Jenkins listening HTTP port
 #
@@ -303,6 +310,7 @@ class jenkins (
   Optional[String] $cli_password_file             = undef,
   Integer $cli_tries                              = 10,
   Integer $cli_try_sleep                          = 10,
+  String $urlbase                                 = 'http://localhost',
   Integer $port                                   = 8080,
   Stdlib::Absolutepath $libdir                    = '/usr/share/java',
   Boolean $manage_datadirs                        = true,
--- a/README.md
+++ b/README.md
@@ -205,9 +205,12 @@ interact with the Jenkins API. Users, Credentials, and security model
 configuration are all driven through this script.
 
 When an API-based resource is defined, the Jenkins' CLI is installed and run
-against the local system (127.0.0.1). Jenkins is assumed to be listening on
+against the local system (localhost). Jenkins is assumed to be listening on
 port 8080, but the module is smart enough to notice if you've configured an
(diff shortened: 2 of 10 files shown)


manifests/init.pp
```diff
@@ -196,6 +196,13 @@
 # @param cli_try_sleep
 #   Seconds between tries to contact jenkins API
 #
+# @param urlbase
+#  Jenkins base url including protocol (http/https) and hostname
+#
+#  Note that this value is use for CLI communication and does not configure your Jenkins system location
```
kenyon: minor typo

```suggestion
#  Note that this value is used for CLI communication and does not configure your Jenkins system location.
```

kenyon (commented):
Small typo, otherwise looks fine.

